PR #2, opened 2020-12-03
makefile: fix build of minimal

### LINKED ISSUE #1: compilation of minimal failing because kernel header path not in includes
On Ubuntu 20.10 this happens to me:

```
$ V=1 make minimal
clang -g -O2 -target bpf -D__TARGET_ARCH_x86 -I.output -c minimal.bpf.c -o .output/minimal.bpf.o
In file included from minimal.bpf.c:3:
In file included from /usr/include/linux/bpf.h:11:
/usr/include/linux/types.h:5:10: fatal error: 'asm/types.h' file not found
#include <asm/types.h>
```

The rub seems to be that `/usr/include/linux/types.h` wants `asm/types.h` which, on my system, I believe lives in `/usr/include/x86_64-linux-gnu/asm/types.h` (which file just reads `#include <asm-generic/types.h>`), and `/usr/include/x86_64-linux-gnu` does not seem to be among clang's include paths, as demonstrated by the `clang -v` output below.

```
clang -v -g -O2 -target bpf -D__TARGET_ARCH_x86 -I.output -c minimal.bpf.c -o .output/minimal.bpf.o
Ubuntu clang version 11.0.0-2
...
#include "..." search starts here:
#include <...> search starts here:
 .output
 /usr/local/include`
 /usr/lib/llvm-11/lib/clang/11.0.0/include
 /usr/include
End of search list.
```

FWIW, on the gcc side, the preprocessor includes `/usr/include/x86_64-linux-gnu` in the `cpp -v` output.

I'm not sure who's at fault here, or what the correct fix is, but adding `-I/usr/include/x86_64-linux-gnu` makes it work. Changing the source to use `vmlinux.h` instead of `<linux/bpf.h>` also makes it work.

Although I don't believe it to be the case, it is possible that my system got screwed up by me manually installing kernel headers through the Debian package created by the kernel's `deb-pkg` make target. I don't think it's the case because I've verified that `/usr/include/linux/types.h` corresponds to the file provided by the Ubuntu's `linux-libc-dev` package.

andreimatei:
Add architecture-specific headers to the BPF compilation command,
otherwise they're missing on some systems (at least on x86-64 Ubuntu).

Fixes #1

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -11,6 +11,17 @@ ARCH := $(shell uname -m | sed 's/x86_64/x86/')
 
 APPS = minimal bootstrap
 
+# Get Clang's default includes on this system. We'll explicitly add these dirs
+# to the includes list when compiling with `-target bpf` because otherwise some
+# architecture-specific dirs will be "missing" on some architectures/distros -
+# headers such as asm/types.h, asm/byteorder.h, asm/socket.h, asm/sockios.h,
+# sys/cdefs.h etc. might be missing.
+#
+# Use '-idirafter': Don't interfere with include mechanics except where the
+# build would have failed anyways.
+CLANG_BPF_SYS_INCLUDES = $(shell $(CLANG) -v -E - </dev/null 2>&1 \
+	| sed -n '/<...> search starts here:/,/End of search list./{ s| \(/.*\)|-idirafter \1|p }')
+
 ifeq ($(V),1)
 	Q =
 	msg =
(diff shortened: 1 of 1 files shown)


src/Makefile
```diff
@@ -11,6 +11,21 @@ ARCH := $(shell uname -m | sed 's/x86_64/x86/')
...
 
+# Get Clang's default includes on this system. We'll explicitly add these dirs
+# to the includes list when compiling with `-target bpf` because otherwise some
+# architecture-specific dirs will be "missing" on some architectures/distros -
+# headers such as asm/types.h, asm/byteorder.h, asm/socket.h, asm/sockios.h,
+# sys/cdefs.h etc. might be missing.
+#
+# Use '-idirafter': Don't interfere with include mechanics except where the
+# build would have failed anyways.
+define get_clang_sys_includes
```
anakryiko, line 22: there is no need to define the function, you can just assign this to CLANG_BPF_SYS_INCLUDES

andreimatei: done, thanks

anakryiko:
Great, thank you!

PR #7, opened 2021-01-20
kprobe: minimal example of doing kprobe/kretprobe BPF programs

jvijtiuk:
Hello,

this PR adds an example that shows how to use kprobes and kretprobes with libbpf, with a description in the README.
The example is similar to the Linux kernel `samples/kprobe/kprobe_example.c` example.

I'm not sure whether fentry/fexit should be mentioned as preferred alternatives in the README?

I've used `%lx` to print the `tls` and `stack` pointers as `%p` prints hashed values. `perf probe` behaves similarly, and prints the whole addresses when it is attached to the `kernel_clone` function and configured to print `stack` and `tls`, so I guess this should be fine for an example. Otherwise, an example that doesn't print addresses could be used, or https://www.kernel.org/doc/html/v4.17/core-api/printk-formats.html could be linked somewhere in the README for clarity.

Regards,
Juraj


src/kprobe.bpf.c
```diff
@@ -0,0 +1,27 @@
...
+#include <bpf/bpf_helpers.h>
+#include <bpf/bpf_tracing.h>
+#include <bpf/bpf_core_read.h>
+
+char LICENSE[] SEC("license") = "Dual BSD/GPL";
+
+SEC("kprobe/kernel_clone")
+int BPF_KPROBE(kernel_clone, struct kernel_clone_args *args)
+{
+	unsigned long stack_size, stack, tls;
```
anakryiko, line 13: this is going to be a problem on 32-bit architectures. unsigned long is always 64-bit for BPF, even on 32-bit architectures. So you'll be reading wrong amount of bytes. It's handled by libbpf automatically for fentry/fexit programs with direct memory accesses, but not for BPF_CORE_READ. Reading longs from kernel is tricky. Is there any other good example you can use?

anakryiko: How about tracking exec() and printing PID and filename of the process? Sort of mini-execsnoop.

src/kprobe.c
```diff
@@ -0,0 +1,65 @@
...
+		fprintf(stderr, "Failed to attach BPF skeleton\n");
+		goto cleanup;
+	}
+
+	printf("Successfully started!\n");
+
+	for (;;) {
+		/* trigger our BPF program */
+		fprintf(stderr, ".");
+		sleep(1);
```
anakryiko, line 59: well, you are not really triggering anything here, unless you are going to do fork() in a loop.

README.md
```diff
@@ -95,6 +95,41 @@ $ sudo cat /sys/kernel/debug/tracing/trace_pipe
...
 
+# Kprobe
+
+`kprobe` is an example of dealing with kernel-space entry and exit (return) probes,
+`kprobe` and `kretprobe` in libbpf lingo. It attaches `kprobe` and `kretprobe`
+BPF programs to the `kernel_clone` function and logs some input arguments
+and return result, respectively, using `bpf_printk()` macro. The example is similar to the
+Linux kernel's `samples/kprobe/kprobe_example.c` example. The example will catch calls to
+`fork`, `clone` and similar functions. Depending on the function called different
+arguments will be set. See the corresponding man pages for more details.
```
anakryiko, line 106: It would indeed be good to call out fentry/fexit as a nicer alternative to kprobe/kretprobe (in terms of usability and performance). Adding fentry demo, that matches kprobe demo would allow to contrast them to each other. So how about doing just that?

src/kprobe.bpf.c
jvijtiuk: Switching to exec turned out to be somewhat more complex than I expected it to be, due to fentry/fexit issues. I'll describe the issues I encountered in a reply to your comment about fentry below. kprobe now tracks bprm_execve and prints the PID and filename. The README also reflects those changes

src/kprobe.c
jvijtiuk: True, I removed it. That was accidentally leftover from minimal.c

README.md
jvijtiuk: I've added a fentry exec() example. However, I'm not sure if it will be the best example due to a few issues. It also attaches to `bprm_execve` so that it matches the kprobe example and to show the ability to use "normal" C in fentry/fexit programs as compared to kprobes. I originally tried to attach to `do_execveat_common`. The alternative was to attach to architecture specific execve syscalls, like `__x64_sys_execveat` which I don't think would make sense for an example. 

`do_execveat_common` worked fine with kprobes, but then I encountered the first issue with fentry. Since libbpf seems to currently use BTF to attach fentry/fexit programs, it couldn't attach to `do_execveat_common` on my system which is running latest Arch Linux kernel. 
The reason for that seems to be that `do_execveat_common` is a static kernel function, and my guess is that the kernel BTF wasn't generated with a recent enough pahole. After manually compiling a 5.10 kernel from source the BTF generated that way included `do_execveat_common`. I assume it works because `do_execveat_common` is listed in the ftrace function list, and newer pahole versions seem to check for that. The error message from libbpf for the vanilla Arch kernel is below:
```
libbpf: do_execveat_common is not found in vmlinux BTF
```

That raises the second issue, I still couldn't attach to `do_execveat_common` with the new kernel because of the following error:
```
The function do_execveat_common arg2 type STRUCT is unsupported.
processed 0 insns (limit 1000000) max_states_per_insn 0 total_states 0 peak_states 0 mark_read 0
```

From what I understand, BTF doesn't support plain struct function arguments, only pointers to structs and `do_execveat_common` has two `struct user_arg_ptr` arguments. Due to all that I decided to hook  `bprm_execve` which does work with fentry/fexit. It will still need kernel BTF generated with recent enough pahole because `bprm_execve` is also a static function in `fs/exec.c`.

At this point I'm not sure whether it would be better to attach to another function, or document the requirements to run this example.

src/fentry.bpf.c
```diff
@@ -0,0 +1,27 @@
...
+	const char *name = filename->name;
+
+	pid = bpf_get_current_pid_tgid();
+
+	bpf_printk("fentry: pid = %d, filename = %s\n", pid, name);
+	return 0;
+}
+
+SEC("fexit/bprm_execve")
+int BPF_PROG(bprm_execve_exit, struct linux_binprm *bprm, int fd, struct filename *filename, int flags, int ret)
```
anakryiko, line 23: can you please wrap this long line to be within 100 lines (kernel standard nowadays)

src/fentry.bpf.c
```diff
@@ -0,0 +1,27 @@
...
+#include <bpf/bpf_helpers.h>
+#include <bpf/bpf_tracing.h>
+#include <bpf/bpf_core_read.h>
+
+char LICENSE[] SEC("license") = "Dual BSD/GPL";
+
+SEC("fentry/bprm_execve")
+int BPF_PROG(bprm_execve, struct linux_binprm *bprm, int fd, struct filename *filename, int flags)
+{
+	pid_t pid;
```
anakryiko, line 13: nit: move it below for "reverse Christmas tree" effect, preferred style in net/bpf kernel trees, thanks!

src/fentry.bpf.c
```diff
@@ -0,0 +1,27 @@
...
+	pid = bpf_get_current_pid_tgid();
+
+	bpf_printk("fentry: pid = %d, filename = %s\n", pid, name);
+	return 0;
+}
+
+SEC("fexit/bprm_execve")
+int BPF_PROG(bprm_execve_exit, struct linux_binprm *bprm, int fd, struct filename *filename, int flags, int ret)
+{
+	bpf_printk("fexit: ret = %d\n", ret);
```
anakryiko, line 25: it would be great to demonstrate that you can actually read input arguments, so please output filename here as well. That's another huge and important difference between kretprobe and fexit (I might have forgotten to emphasize that, please add to the README as well). This is a critical feature in some cases.

src/fentry.c
```diff
@@ -0,0 +1,64 @@
...
+	if (err) {
+		fprintf(stderr, "Failed to attach BPF skeleton\n");
+		goto cleanup;
+	}
+
+	printf("Successfully started!\n");
+
+	for (;;) {
+		fprintf(stderr, ".");
+		sleep(1);
```
anakryiko, line 58: There is little point to do this loop. How about removing the loop and enhancing above printf like so:

```c
printf("Successfully started! Please run `sudo cat /sys/kernel/debug/tracing/trace_pipe` to see output of the BPF programs.\n");
```

Similarly for kprobe example.

src/kprobe.bpf.c
```diff
@@ -0,0 +1,27 @@
...
+#include <bpf/bpf_helpers.h>
+#include <bpf/bpf_tracing.h>
+#include <bpf/bpf_core_read.h>
+
+char LICENSE[] SEC("license") = "Dual BSD/GPL";
+
+SEC("kprobe/bprm_execve")
+int BPF_KPROBE(bprm_execve, struct linux_binprm *bprm, int fd, struct filename *filename, int flags)
+{
+	pid_t pid;
```
anakryiko, line 13: same, please move one line below, it looks a bit weird

anakryiko (commented):
I think `bprm_execve` is ok, just mention that you need `pahole` v1.19+ to be able to attach to static kernel functions with fentry/fexit. In a year this won't be an issue at all.

jvijtiuk:
Thanks for the feedback. I've fixed everything except removing the for loops, as the programs immediately exit without a loop. I have however replaced them with loops that run until SIGINT is received, similar to some BCC libbpf-tools examples. Additionally cleanup is always explicitly done that way.

The fexit example now prints the filename and PID so that the output matches the fentry one.

src/fentry.bpf.c
```diff
@@ -0,0 +1,32 @@
+// SPDX-License-Identifier: GPL-2.0 OR BSD-3-Clause
+/* Copyright (c) 2021 Sartura */
+#include "vmlinux.h"
+#include <bpf/bpf_helpers.h>
+#include <bpf/bpf_tracing.h>
+#include <bpf/bpf_core_read.h>
```
anakryiko, line 6: you don't really use this header, can you please remove it?

README.md
```diff
@@ -95,6 +95,75 @@ $ sudo cat /sys/kernel/debug/tracing/trace_pipe
...
+```
+
+# Fentry
+`fentry` is an example that uses fentry and fexit tracing. It attaches `fentry` and `fexit` traces to `bprm_execve`
+which is called by the execve family of functions and logs the return value, PID and filename to the trace pipe.
+Important differences compared to kprobes are improved performance and usability. fentry and fexit programs are available
+starting from 5.5 kernels.
+In this example easier usability is shown with the ability to directly dereference pointer arguments like in normal C,
+instead of using various read helpers.
+The second difference is the requirement to list all the function arguments, followed by the return value,
```
anakryiko, line 135: it's not really a requirement, you can do just:

int my_fentry(void *ctx) { ... }

and it will work just fine. So would be good to mild it down this sentence a bit

anakryiko (commented):
Looks great with two minor points. Can you please squash all your commits into either one or two (one for fentry one for kprobe) commits, so that we maintain a clean history? Thanks!

anakryiko:
Thanks for the contribution! I took some liberty in rewording fentry/fexit description to reduce possible confusion. And did some slight adjustments to code formating here and there. I hope that's fine.

One thing I noticed, though, is that bprm_execve() is a pretty recent change in kernel, so those examples will only work on very recent kernels. Which is quite unfortunate. If you can, please think about other simple, but interesting examples that could server as a demo for kprobe/fentry, maybe we can update them to work on older kernels just fine.

jvijtiuk:
Sorry, I jut saw this today. I was worried about bprm_execve() as well. I'm currently looking for an alternative and will comment again here when I find something usable.

PR #10, opened 2021-02-07
Update gen_vmlinux_h.sh

Vesnica:
Positional or special parameters cannot be assigned using `:=` way. Should be `:-`.


anakryiko:
You are right, thank you!

PR #11, opened 2021-02-07
fentry, kprobe: attach fentry and kprobe examples to do_unlinkat()

jvijtiuk:
As suggested in #7, this changes the fentry and kprobe examples so that now they hook do_unlinkat() instead of bprm_execve().

I hope this will be fine as the function hasn't been changed in a while, is easy to trigger, and most of the example remains the same.

I've updated the README, and removed the part about pahole 1.19, as this function isn't static.


anakryiko (approved):
This is great, thanks!

PR #18, opened 2021-04-10
Add XDP/Rust example

chenhengqi:
This PR adds a XDP example using [libbpf-rs](https://github.com/libbpf/libbpf-rs).


chenhengqi:
cc @danobi

danobi:
Wrote down some thoughts on directory structure here: https://github.com/libbpf/libbpf-bootstrap/issues/19

chenhengqi:
OK. Let's continue the discussion in the issue.

src/kprobe.bpf.c
```diff
@@ -10,18 +10,21 @@ char LICENSE[] SEC("license") = "Dual BSD/GPL";
...
 
 SEC("kretprobe/do_unlinkat")
-int BPF_KRETPROBE(do_open_execat_exit, long ret)
+int BPF_KRETPROBE(do_unlinkat_exit, long ret)
 {
-	bpf_printk("KPROBE EXIT: ret = %ld\n", ret);
+	pid_t pid;
+
+	pid = bpf_get_current_pid_tgid() >> 32;
+	bpf_printk("KPROBE EXIT: pid=%d, ret = %ld\n", pid, ret);
```
anakryiko, line 28: nit: `pid = %d` ?

src/xdpdrop/src/main.rs
```diff
@@ -0,0 +1,47 @@
...
+
+    let skel_builder = XdpdropSkelBuilder::default();
+    let open_skel = skel_builder.open()?;
+    let mut skel = open_skel.load()?;
+    let link = skel.progs().xdp_drop().attach_xdp(opts.ifindex)?;
+    skel.links = XdpdropLinks{
+        xdp_drop: Some(link),
+    };
+
+    thread::sleep(time::Duration::from_millis(opts.duration));
```
anakryiko, line 45: can we wait for ctrl-c instead of timeout?

src/xdpdrop/src/bpf/vmlinux.h
```diff
@@ -0,0 +1 @@
+../../../vmlinux_508.h
```
anakryiko, line 1: should probably point to vmlinux.h (which is a symlink to specific version). But ideally, is it possible specify location of vmlinux.h somehow without this?

anakryiko (commented):
*facepalm* about usdt... I'm inclined to rewrite the history to reduce the size of the repository :(

src/xdpdrop/src/bpf/vmlinux.h
chenhengqi: Maybe we should add a `-I` option to libbpf-rs's `cargo libbpf build` command.

PR #20, opened 2021-04-14
Add XDP/Rust example

chenhengqi:
Replace #18


danobi:
@chenhengqi #21 is in now. Can you move the rust example into `examples/rust`?

chenhengqi:
@danobi  Done.

cc @anakryiko

examples/rust/xdp/Cargo.toml
```diff
@@ -0,0 +1,14 @@
...
+authors = ["Hengqi Chen <chenhengqi@outlook.com>"]
+edition = "2018"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+anyhow = "1.0"
+ctrlc = { version = "3.0", features = ["termination"] }
+libc = "0.2"
+libbpf-rs = { git = "https://github.com/libbpf/libbpf-rs" }
```
danobi, line 13: I've released a version with your XDP patch in -- can you pin the version instead? So that any future API breaking releases don't automatically break these examples as well.

examples/rust/xdp/src/main.rs
```diff
@@ -0,0 +1,57 @@
...
+    skel.links = XdpdropLinks {
+        xdp_drop: Some(link),
+    };
+
+    let running = Arc::new(AtomicBool::new(true));
+    let r = running.clone();
+    ctrlc::set_handler(move || {
+        r.store(false, Ordering::SeqCst);
+    })
+    .expect("Error setting Ctrl-C handler");
```
danobi, line 49: Could just:
```suggestion
    })?;
```

or if the custom error is really necessary:

```
.with_context(|| format!("Error setting Ctrl-C handler"))?;
```

examples/rust/xdp/src/main.rs
```diff
@@ -0,0 +1,57 @@
...
+
+    while running.load(Ordering::SeqCst) {
+        eprint!(".");
```
danobi, line 52: Why the `.`?

danobi (commented):
Curious: does running this mess up the networking on your computer?

examples/rust/xdp/src/main.rs
chenhengqi: Ah, just like the c examples do. Any suggestion ?

chenhengqi: Ah, just like the c examples do. Any suggestion ?

examples/rust/xdp/src/main.rs
chenhengqi: Yes, the error propagation way is good enough for this simple example.

chenhengqi:
> Curious: does running this mess up the networking on your computer?

I work with Docker bridge network, so nope.

examples/rust/xdp/src/main.rs
danobi: Oh, that's fine the way it is then.

danobi:
LGTM

chenhengqi:
@anakryiko  Can we get this landed ?

anakryiko:
Oops, sorry, fell off my plate. Can you please follow up with the update to README mentioning that there are Rust examples and how to build/run them? Thanks!

Issue #19, opened 2021-04-12
Support multiple languages / platforms


danobi:
Creating this issue to decide directory structure and code layout.

Current structure:
```
.
├── libbpf
├── LICENSE
├── README.md
├── src
│   ├── bootstrap.bpf.c
│   ├── bootstrap.c
│   ├── bootstrap.h
│   ├── fentry.bpf.c
│   ├── fentry.c
│   ├── kprobe.bpf.c
│   ├── kprobe.c
│   ├── Makefile
│   ├── minimal.bpf.c
│   ├── minimal.c
│   ├── uprobe.bpf.c
│   ├── uprobe.c
│   ├── vmlinux_508.h
│   └── vmlinux.h -> vmlinux_508.h
└── tools
    ├── bpftool
    └── gen_vmlinux_h.sh
```

Proposed structure:
```
.
├── examples
│   ├── c
│   │   ├── bootstrap
│   │   │   ├── bootstrap.bpf.c
│   │   │   ├── bootstrap.c
│   │   │   ├── bootstrap.h
│   │   │   └── Makefile
│   │   ├── fentry
│   │   │   ├── fentry.bpf.c
│   │   │   ├── fentry.c
│   │   │   └── Makefile
│   │   ├── kprobe
│   │   │   ├── kprobe.bpf.c
│   │   │   ├── kprobe.c
│   │   │   └── Makefile
│   │   ├── Makefile
│   │   ├── minimal
│   │   │   ├── Makefile
│   │   │   ├── minimal.bpf.c
│   │   │   └── minimal.c
│   │   └── uprobe
│   │       ├── Makefile
│   │       ├── uprobe.bpf.c
│   │       └── uprobe.c
│   └── rust
│       ├── Cargo.toml
│       ├── Makefile
│       └── xdp
│           ├── Cargo.toml
│           ├── Makefile
│           └── src
│               ├── bpf
│               │   └── xdp.bpf.c
│               └── main.rs
├── libbpf
├── LICENSE
├── README.md
├── tools
│   ├── bpftool
│   └── gen_vmlinux_h.sh
└── vmlinux
    ├── Makefile
    ├── vmlinux_508.h
    └── vmlinux.h -> vmlinux_508.h
```

Basic idea is top level entries (`tools/`, `vmlinux/`, etc.) are shared between _all_ examples. Anything language / platform specific must be completely contained to the `examples/*` directory. Each language/platform has a makefile that does the right thing.

[...]

PR #21, opened 2021-04-14
Reorganize C examples 

danobi:
This reorganizes the examples to make room for other languages and
platforms as described in #19 .

There are now Makefiles for each individual example that delegate to the
parent Makefile. This makes it possible to type `make` anywhere and get
what you expect.


examples/c/Makefile.individual
```diff
@@ -0,0 +1,10 @@
+BINARY := $(notdir $(realpath .))
+TARGET := $(BINARY)/$(BINARY)
+
+.PHONY: all
+all:
+	@make --no-print-directory -C .. $(TARGET)
```
anakryiko, line 6: You should use `$(MAKE)` instead, there is a special section(s) covering this in Make manual :)

examples/c/Makefile.individual
```diff
@@ -0,0 +1,10 @@
...
+
+.PHONY: clean
+clean:
+	rm $(BINARY)
```
anakryiko, line 10: well this is not exactly cleaning all the relevant files, actually. $(OUTPUT)/$(BINARY) should be cleaned at least. We can add `make clean-$(BINARY)` targets for each known app, that would be nicer, probably.

examples/c/Makefile
```diff
@@ -2,17 +2,18 @@
...
-LIBBPF_SRC := $(abspath ../libbpf/src)
+BPFTOOL ?= $(abspath ../../tools/bpftool)
+LIBBPF_SRC := $(abspath ../../libbpf/src)
 LIBBPF_OBJ := $(abspath $(OUTPUT)/libbpf.a)
+VMLINUX := ../../vmlinux/vmlinux.h
 # Use our own libbpf API headers and Linux UAPI headers distributed with
 # libbpf to avoid dependency on system-wide headers, which could be missing or
 # outdated
-INCLUDES := -I$(OUTPUT) -I../libbpf/include/uapi
+INCLUDES := -I$(OUTPUT) -I../../libbpf/include/uapi -I$(dir $(VMLINUX))
```
anakryiko, line 12: I think we needed -I$(OUTPUT) for skeletons, but now they are in their app's directory, so this probably can be dropped?

examples/c/Makefile
```diff
@@ -58,8 +59,9 @@ $(LIBBPF_OBJ): $(wildcard $(LIBBPF_SRC)/*.[ch] $(LIBBPF_SRC)/Makefile) | $(OUTPU
 		    install
 
 # Build BPF code
-$(OUTPUT)/%.bpf.o: %.bpf.c $(LIBBPF_OBJ) $(wildcard %.h) vmlinux.h | $(OUTPUT)
+$(OUTPUT)/%.bpf.o: %.bpf.c $(LIBBPF_OBJ) $(wildcard %.h) $(VMLINUX) | $(OUTPUT)
```
anakryiko, line 62: hm... did you try adding `$(dir $@)` or `$(dir $(OUTPUT)/%)`, just curious. It would be nice to have a nice MKDIR message, similarly to $(OUTPUT) and $(OUTPUT)/libbpf cases

examples/c/Makefile
```diff
@@ -2,17 +2,18 @@
...
 # Use our own libbpf API headers and Linux UAPI headers distributed with
 # libbpf to avoid dependency on system-wide headers, which could be missing or
 # outdated
-INCLUDES := -I$(OUTPUT) -I../libbpf/include/uapi
+INCLUDES := -I$(OUTPUT) -I../../libbpf/include/uapi -I$(dir $(VMLINUX))
 CFLAGS := -g -Wall
 ARCH := $(shell uname -m | sed 's/x86_64/x86/')
 
-APPS = minimal bootstrap uprobe kprobe fentry
+APPS = minimal/minimal bootstrap/bootstrap uprobe/uprobe kprobe/kprobe fentry/fentry
```
anakryiko, line 16: also, let's keep it as just `minimal bootstrap etc`, so that `make minimal` would do what you'd expect.

anakryiko (commented):
Some of the Makefile suggestions can be addressed separately, if you'll run into troubles figuring it out. Just let me know.

Thanks for working on this!

examples/c/Makefile.individual
chenhengqi: I think it would be better to add a `-f` flag to `rm`.

examples/c/bootstrap/Makefile
```diff
@@ -0,0 +1 @@
+../Makefile.individual
```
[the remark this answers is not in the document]
danobi: It is symlinked -- "../Make..." is just github's way of displaying it

examples/c/Makefile
danobi: Seems it's still needed for installed libbpf headers

examples/c/Makefile.individual
danobi: Ok

examples/c/Makefile
danobi: Chatted offline -- will just hardcode for now. Strangely complex to do this in a makefile.

examples/c/Makefile
danobi: Done

anakryiko (changes requested):
This got very ugly fast... Let me think a bit more about this and get back to this. The point was to demonstrate how *simple* real-world Makefile would be set up to compile an application, and instead we get this advanced monstrosity of a Makefile...

danobi:
> This got very ugly fast... Let me think a bit more about this and get back to this. The point was to demonstrate how _simple_ real-world Makefile would be set up to compile an application, and instead we get this advanced monstrosity of a Makefile...

I'm finishing up the cmake build right now FWIW. TBH I'm not sure many project maintain makefiles by hand these days. IMHO the makefile is not very realistic to begin with.

danobi:
Here's what the cmake build looks like: https://github.com/danobi/libbpf-bootstrap/commit/4df9c5a4259573a0298593fbacffac76e9a03d5c

anakryiko (approved):
Awesome, thanks!

PR #25, opened 2021-05-06
Update XDP example and README

chenhengqi:
1. Update the XDP example to log packet size and return `XDP_PASS`, so that the program won't mess up the network and user can have observability through trace_pipe.
2. Update README, add a section for the XDP program, add a section for Rust building. 
    BTW, switch the position of `kprobe`/`fentry` sections since the `fentry` section mentions `kprobe`.


anakryiko:
Let's keep fentry section first as is, as it's the recommended way.

chenhengqi:
OK, updated.

anakryiko:
Great, thanks!

PR #29, opened 2021-05-30
rust: xdppass: Use libbpf-cargo build script

danobi:
Build script is better than cargo plugin b/c you can't forget to run the
skeleton generation step.


danobi:
cc @chenhengqi

chenhengqi:
It looks good.

One small question, the current implementation use relative path rather than `CARGO_MANIFEST_DIR`, does it works in both workspace dir and crate dir?

danobi:
> One small question, the current implementation use relative path rather than CARGO_MANIFEST_DIR, does it works in both workspace dir and crate dir?

It should work. https://doc.rust-lang.org/cargo/reference/build-scripts.html says:

> In addition to environment variables, the build script’s current directory is the source directory of the build script’s package.

danobi:
@anakryiko  ping

PR #30, opened 2021-06-03
Add example for uprobes and kprobes

mkulke:
Added a sample application for libbpf-rs. It will output all TCPv4 connections that have been established on the host as ips and hostnames by probing `tcp_v4_connect` in kernel and glibc's `getaddrinfo` in userland. On a successful host lookup the first result will be stored in a hashmap, which can be used as a lookup table to retrieve a hostname for ip_v4 connections. When the program is started there should be console output when TCPv4 connections are established on the system, eg. via curl:

```bash   
host event: www.jsonplaceholder.com
ip event: 172.67.201.157
```

### DIFF
--- a/examples/rust/Cargo.toml
+++ b/examples/rust/Cargo.toml
@@ -1,4 +1,5 @@
 [workspace]
 members = [
+  "tracecon",
   "xdp",
 ]
--- a/examples/rust/tracecon/README.md
+++ b/examples/rust/tracecon/README.md
@@ -1,0 +1,88 @@
+# tracecon
+
+An eBPF sample application, written in C & Rust using
+[libbpf-rs](https://github.com/libbpf/libbpf-rs). It will output all
+TCPv4 connections that have been established on the host as ips and
+hostnames by probing `tcp_v4_connect` in kernel and glibc's `getaddrinfo`
+in userland. On a successful host lookup the first result will be stored in
+a hashmap, which can be used as a lookup table to retrieve a hostname for
+ip_v4 connections.
+
+## Requirements
+
+### Kernel
+
+The project is built on technology like `CO-RE` and `BTF`, which is only
+available in more recent kernels (5.0-ish). Ubuntu 20.10 has configured and
+packaged all the required dependencies.
+
+### Compilers
+
+The project has been tested with LLVM v11 and Rust v1.52.1.
+
+### Generate `vmlinux.h`
+
+```bash
+bpftool btf dump file /sys/kernel/btf/vmlinux format c > src/bpf/vmlinux.h
+```
+
+You can verify whether your kernel was built with BTF enabled:
+
+```bash
+cat /boot/config-$(uname -a) | grep CONFIG_DEBUG_INFO_BTF
+```
+
+## Build
+
+### Vagrant
+
+eBPF is a low-level technology on the Linux kernel. Docker is not a good fit
+to build eBPF code on MacOS or Windows environments. On those platforms
+Docker ships its own kernel (e.g. linuxkit) and BTF might not be enabled.
+
+There is a `Vagrantfile` to provision a Ubuntu 20.10 VM including the
+necessary dependencies to build the project. To install Vagrant with a
+VirtualBox backend and provision the VM on a MacOS host machine run:
+
+```
+brew cask install virtualbox
+brew cask install vagrant
+vagrant up
+```
+
+Log in to the machine. The current host workdir is mounted to `/vagrant`:
+
+```
+vagrant ssh
+sudo su -
+cd /vagrant
+```
+
+### Cargo
+
+```bash
+cargo build
+```
+
+## Run
+
+Start the program to instrument the eBPF probe and listen to events:
+
+```bash
+cargo run --release
+```
+
+In another shell perform some http calls:
+
+```bash
+curl -s www.jsonplaceholder.com > /dev/null
+# Do not use a dns lookup
+curl -s -H "Host: www.jsonplaceholder.com" 172.67.201.157 > /dev/null
+```
+
+The other shell should show the respective events:
+
+```bash
+host event: www.jsonplaceholder.com
+ip event: 172.67.201.157
+```
--- a/examples/rust/tracecon/src/bpf/tracecon.bpf.c
+++ b/examples/rust/tracecon/src/bpf/tracecon.bpf.c
@@ -1,0 +1,162 @@
+// SPDX-License-Identifier: GPL-2.0 OR BSD-3-Clause
+#include "vmlinux.h"
+#include <bpf/bpf_helpers.h>
+#include <bpf/bpf_core_read.h>
+#include <bpf/bpf_tracing.h>
+
+#define memcpy(dest, src, n) __builtin_memcpy((dest), (src), (n))
+#define HOSTNAME_LEN 84
+
+const volatile pid_t target_pid = 0;
+
+/* Copied from: include/netdb.h */
+struct addrinfo {
+	int ai_flags; /* Input flags.  */
+	int ai_family; /* Protocol family for socket.  */
+	int ai_socktype; /* Socket type.  */
+	int ai_protocol; /* Protocol for socket.  */
+	u32 ai_addrlen; /* Length of socket address.  */
+	struct sockaddr *ai_addr; /* Socket address for socket.  */
+	char *ai_canonname; /* Canonical name for service location.  */
+	struct addrinfo *ai_next; /* Pointer to next in list.  */
+};
+
+struct lookup {
+	char c[84];
+	struct addrinfo **results;
+};
+
+struct {
+	__uint(type, BPF_MAP_TYPE_HASH);
+	__uint(max_entries, 10240);
+	__type(key, u32);
+	__type(value, struct lookup);
+} lookups SEC(".maps");
+
+struct {
+	__uint(type, BPF_MAP_TYPE_HASH);
+	__uint(max_entries, 10240);
+	__type(key, u32);
+	__type(value, struct lookup);
+} hostnames SEC(".maps");
+
+struct {
+	__uint(type, BPF_MAP_TYPE_HASH);
+	__uint(max_entries, 10240);
+	__type(key, u32);
+	__type(value, struct sock *);
+} sockets SEC(".maps");
+
+struct {
+	__uint(type, BPF_MAP_TYPE_PERF_EVENT_ARRAY);
+	__uint(key_size, sizeof(u32));
+	__uint(value_size, sizeof(u32));
+} events SEC(".maps");
+
+enum tag { IP = 0, HOSTNAME = 1 };
+
+struct event {
+	u8 tag;
+	u8 ip[4];
+	u8 hostname[HOSTNAME_LEN];
+};
+
+/* trigger creation of event struct in skeleton code */
+struct event _event = {};
+
+static u32 get_tid()
+{
+	u64 tgid = bpf_get_current_pid_tgid();
+	pid_t pid = tgid >> 32;
+
+	if (target_pid != 0 && target_pid != pid)
+		return 0;
+	return (u32)tgid;
+}
+
+SEC("uprobe/getaddrinfo")
+int BPF_KPROBE(getaddrinfo_enter, const char *hostname, const char *service,
+	       const struct addrinfo *hints, struct addrinfo **res)
+{
+	u32 tid = get_tid();
+	struct lookup lookup = {};
+
+	if (!tid)
+		return 0;
+	bpf_probe_read_user_str(&lookup.c, sizeof(lookup.c), hostname);
+	lookup.results = res;
+	bpf_map_update_elem(&lookups, &tid, &lookup, BPF_ANY);
+	return 0;
+}
+
+SEC("uretprobe/getaddrinfo")
+int BPF_KRETPROBE(getaddrinfo_exit, int ret)
+{
+	u32 tid = get_tid();
+	struct lookup *lookup;
+	struct addrinfo *result;
+	struct sockaddr_in *addr;
+	struct in_addr ipv4_addr;
+
+	if (!tid)
+		return 0;
+	if (ret != 0)
+		goto cleanup;
+	lookup = bpf_map_lookup_elem(&lookups, &tid);
+	if (!lookup)
+		return 0;
+	bpf_probe_read_user(&result, sizeof(result), lookup->results);
+	bpf_probe_read_user(&addr, sizeof(addr), &result->ai_addr);
+	bpf_probe_read_user(&ipv4_addr, sizeof(ipv4_addr), &addr->sin_addr);
+	bpf_map_update_elem(&hostnames, &ipv4_addr.s_addr, lookup, BPF_ANY);
+cleanup:
+	bpf_map_delete_elem(&lookups, &tid);
+	return 0;
+}
+
+
+SEC("kprobe/tcp_v4_connect")
+int BPF_KPROBE(tcp_v4_connect_enter, struct sock *sk, struct sockaddr *uaddr, int addr_len)
+{
+	u32 tid = get_tid();
+
+	if (!tid)
+		return 0;
+	bpf_map_update_elem(&sockets, &tid, &sk, 0);
+	return 0;
+};
+
+SEC("kretprobe/tcp_v4_connect")
+int BPF_KRETPROBE(tcp_v4_connect_exit, int ret)
+{
+	u32 tid = get_tid();
+	struct sock **sockpp;
+	struct lookup *lookup;
+	struct event event = {};
+	u32 ip;
+
+	if (!tid)
+		return 0;
+	if (ret != 0)
+		goto cleanup;
+	sockpp = bpf_map_lookup_elem(&sockets, &tid);
+	if (!sockpp)
+		return 0;
+	ip = BPF_CORE_READ(*sockpp, __sk_common.skc_daddr);
+	lookup = bpf_map_lookup_elem(&hostnames, &ip);
+	if (!lookup) {
+		event.tag = IP;
+		memcpy(&event.ip, &ip, sizeof(event.ip));
+	} else {
+		event.tag = HOSTNAME;
+		memcpy(&event.hostname, &lookup->c, sizeof(lookup->c));
+		bpf_map_delete_elem(&hostnames, &ip);
+	}
+	/* ctx is implied in the signature macro */
+	bpf_perf_event_output(ctx, &events, BPF_F_CURRENT_CPU, &event, sizeof(event));
+cleanup:
+	bpf_map_delete_elem(&sockets, &tid);
+	return 0;
+}
+
+char LICENSE[] SEC("license") = "Dual BSD/GPL";
--- a/examples/rust/tracecon/src/main.rs
+++ b/examples/rust/tracecon/src/main.rs
@@ -1,0 +1,127 @@
+use anyhow::{anyhow, bail, Result};
+use core::time::Duration;
+use libbpf_rs::PerfBufferBuilder;
+use object::Object;
+use object::ObjectSymbol;
+use plain::Plain;
+use std::fs;
+use std::net::Ipv4Addr;
+use std::path::Path;
+use std::sync::atomic::{AtomicBool, Ordering};
+use std::sync::Arc;
+use structopt::StructOpt;
+
+#[path = "bpf/.output/tracecon.skel.rs"]
+mod tracecon;
+use tracecon::*;
+
+type Event = tracecon_bss_types::event;
+unsafe impl Plain for Event {}
+
+#[derive(Debug, StructOpt)]
+struct Command {
+    /// verbose output
+    #[structopt(long, short)]
+    verbose: bool,
+    /// glibc path
+    #[structopt(long, short, default_value = "/lib/x86_64-linux-gnu/libc.so.6")]
+    glibc: String,
+    #[structopt(long, short)]
+    /// pid to observe
+    pid: Option<i32>,
+}
+
+fn bump_memlock_rlimit() -> Result<()> {
+    let rlimit = libc::rlimit {
+        rlim_cur: 128 << 20,
+        rlim_max: 128 << 20,
+    };
+
+    if unsafe { libc::setrlimit(libc::RLIMIT_MEMLOCK, &rlimit) } != 0 {
+        bail!("Failed to increase rlimit");
+    }
+
+    Ok(())
+}
+
+fn get_symbol_address(so_path: &str, fn_name: &str) -> Result<usize> {
+    let path = Path::new(so_path);
+    let buffer = fs::read(path)?;
+    let file = object::File::parse(buffer.as_slice())?;
+
+    let mut symbols = file.dynamic_symbols();
+    let symbol = symbols
+        .find(|symbol| {
+            if let Ok(name) = symbol.name() {
+                return name == fn_name;
+            }
+            false
+        })
+        .ok_or(anyhow!("symbol not found"))?;
+
+    Ok(symbol.address() as usize)
+}
+
+fn handle_event(_cpu: i32, data: &[u8]) {
+    let mut event = Event::default();
+    plain::copy_from_bytes(&mut event, data).expect("Event data buffer was too short");
+
+    match event.tag {
+        0 => println!("ip event: {}", Ipv4Addr::from(event.ip)),
+        1 => println!("host event: {}", String::from_utf8_lossy(&event.hostname)),
+        _ => {}
+    }
+}
+
+fn main() -> Result<()> {
+    let opts = Command::from_args();
+
+    let mut skel_builder = TraceconSkelBuilder::default();
+    if opts.verbose {
+        skel_builder.obj_builder.debug(true);
+    }
+
+    bump_memlock_rlimit()?;
+    let mut open_skel = skel_builder.open()?;
+    if let Some(pid) = opts.pid {
+        open_skel.rodata().target_pid = pid;
+    }
+    let mut skel = open_skel.load()?;
+    let address = get_symbol_address(&opts.glibc, "getaddrinfo")?;
+
+    let _uprobe =
+        skel.progs_mut()
+            .getaddrinfo_enter()
+            .attach_uprobe(false, -1, &opts.glibc, address)?;
+
+    let _uretprobe =
+        skel.progs_mut()
+            .getaddrinfo_exit()
+            .attach_uprobe(true, -1, &opts.glibc, address)?;
+
+    let _kprobe = skel
+        .progs_mut()
+        .tcp_v4_connect_enter()
+        .attach_kprobe(false, "tcp_v4_connect")?;
+
+    let _kretprobe = skel
+        .progs_mut()
+        .tcp_v4_connect_exit()
+        .attach_kprobe(true, "tcp_v4_connect")?;
+
+    let perf = PerfBufferBuilder::new(skel.maps_mut().events())
+        .sample_cb(handle_event)
+        .build()?;
+
+    let running = Arc::new(AtomicBool::new(true));
+    let r = running.clone();
+    ctrlc::set_handler(move || {
+        r.store(false, Ordering::SeqCst);
+    })?;
+
+    while running.load(Ordering::SeqCst) {
+        perf.poll(Duration::from_millis(100))?;
+    }
+
+    Ok(())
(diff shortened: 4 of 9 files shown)


examples/rust/tracecon/src/bpf/tracecon.bpf.c
```diff
@@ -0,0 +1,134 @@
...
+	int ai_family;        		/* Protocol family for socket.  */
+	int ai_socktype;      		/* Socket type.  */
+	int ai_protocol;      		/* Protocol for socket.  */
+	u32 ai_addrlen;       		/* Length of socket address.  */ // CHANGED from socklen_t
+	struct sockaddr *ai_addr; 	/* Socket address for socket.  */
+	char *ai_canonname;       	/* Canonical name for service location.  */
+	struct addrinfo *ai_next; 	/* Pointer to next in list.  */
+};
+
+struct lookup_T {
```
danobi, line 21: the _T suffix is a POSIX thing -- `lookup` should be fine alone
```suggestion
struct lookup {
```

examples/rust/tracecon/src/bpf/tracecon.bpf.c
```diff
@@ -0,0 +1,134 @@
...
+
+struct {
+	__uint(type, BPF_MAP_TYPE_PERF_EVENT_ARRAY);
+	__uint(key_size, sizeof(u32));
+	__uint(value_size, sizeof(u32));
+} events SEC(".maps");
+
+enum tag_T { IP = 0, HOSTNAME = 1};
+
+struct event_T {
```
danobi, line 55: ditto

examples/rust/tracecon/src/bpf/tracecon.bpf.c
```diff
@@ -0,0 +1,134 @@
...
+}
+
+SEC("uretprobe/getaddrinfo")
+int BPF_KRETPROBE(getaddrinfo_exit, int ret)
+{
+	u32 tid = bpf_get_current_pid_tgid();
+	struct lookup_T *lookup = bpf_map_lookup_elem(&lookups, &tid);
+	if (lookup == 0) {
+        	return 0;
+        }
```
danobi, line 79:
```suggestion
	if (!lookup)
        	return 0;
```

examples/rust/tracecon/src/bpf/tracecon.bpf.c
```diff
@@ -0,0 +1,134 @@
...
+	return 0;
+};
+
+SEC("kretprobe/tcp_v4_connect")
+int BPF_KRETPROBE(tcp_v4_connect_exit, int ret) {
+	u32 tid = bpf_get_current_pid_tgid();
+	struct sock **sockpp = bpf_map_lookup_elem(&sockets, &tid);
+	if (sockpp == 0) {
+		return 0;
+	}
```
danobi, line 109: Same as above

examples/rust/tracecon/src/bpf/tracecon.bpf.c
```diff
@@ -0,0 +1,134 @@
...
+        }
+	bpf_map_delete_elem(&lookups, &tid);
+	return 0;
+}
+
+SEC("kprobe/tcp_v4_connect")
+int BPF_KPROBE(tcp_v4_connect_enter, struct sock *sk, struct sockaddr *uaddr, int addr_len)
+{
+	u32 tid = bpf_get_current_pid_tgid();
+	bpf_map_update_elem(&sockets, &tid, &sk, 0);
```
danobi, line 99: Why store ptr to ptr and not just `sk`? `sk` is a local -- not sure we can save addr to a local.

examples/rust/tracecon/src/bpf/tracecon.bpf.c
```diff
@@ -0,0 +1,134 @@
...
+		if (lookup == 0) {
+			event.tag = IP;
+			event.ip = daddr;
+		}
+		else {
+			event.tag = HOSTNAME;
+			memcpy(&event.hostname, &lookup->c, sizeof(lookup->c));
+			bpf_map_delete_elem(&hostnames, &daddr);
+		}
+		// ctx is implied in the signature macro
```
danobi, line 127: No c++ style comments in kernel style guide
```suggestion
		/* ctx is implied in the signature macro */
```

examples/rust/tracecon/src/main.rs
```diff
@@ -0,0 +1,121 @@
...
+mod tracecon;
+use tracecon::*;
+
+#[derive(Debug, StructOpt)]
+struct Command {
+    /// verbose output
+    #[structopt(long, short)]
+    verbose: bool,
+    /// glibc path
+    #[structopt(long, short, default_value = "/lib/x86_64-linux-gnu/libc.so.6")]
```
danobi, line 24: FWIW: in libbpf-tools (bcc repo) we've suggested parsing `/proc/self/maps` to get path to active libc. But may be out of scope for this example tool.

examples/rust/tracecon/src/main.rs
```diff
@@ -0,0 +1,121 @@
...
+
+    Ok(symbol.address() as usize)
+}
+
+fn handle_event(_cpu: i32, data: &[u8]) {
+    // payload: 4 bytes tag + 4 bytes ip + 84 bytes hostname
+    let tag = data[0];
+    if tag == 0 {
+        let ip_bytes: [u8; 4] = data[4..8].try_into().unwrap();
+        let ip: Ipv4Addr = ip_bytes.into();
```
danobi, line 64: https://doc.rust-lang.org/stable/std/net/struct.Ipv4Addr.html doesn't mention anything about internal layout. It also mentions:

> The size of an Ipv4Addr struct may vary depending on the target operating system.

Use constructor instead? https://doc.rust-lang.org/stable/std/net/struct.Ipv4Addr.html#method.new

examples/rust/tracecon/src/main.rs
```diff
@@ -0,0 +1,121 @@
...
+            }
+            false
+        })
+        .ok_or(anyhow!("symbol not found"))?;
+
+    Ok(symbol.address() as usize)
+}
+
+fn handle_event(_cpu: i32, data: &[u8]) {
+    // payload: 4 bytes tag + 4 bytes ip + 84 bytes hostname
```
danobi, line 60: Hardcoding offsets isn't very nice. Maybe try using a global in the .bpf.c to trigger definition in skeleton. And then use generated type?

eg what runqslower does: https://github.com/libbpf/libbpf-rs/blob/4b5ee8dc147e781c4ba237efc7108c0f910b1963/examples/runqslower/src/bpf/runqslower.bpf.c#L13-L14

examples/rust/tracecon/src/main.rs
danobi: Oh nevermind, you're using `From` trait. Ignore this

examples/rust/tracecon/src/bpf/tracecon.bpf.c
mkulke: hmm, not sure I understand: the map's value is indeed `struct sock*`. We cannot load `sock` locally, since `sizeof(struct sock)` will exceed the BPF stack size, and I suppose the struct is not populated at this point. hence the argument to the fn is passed from the enter to the exit probe, so we can read it when the connection has been established. Is there a better way to achieve that?

examples/rust/tracecon/src/main.rs
mkulke: That's interesting, would have to dig into that that though and submit it at a later point. I also noticed in bcc you could attach uprobes using merely a `libc` identifier, I will look into this code.

examples/rust/tracecon/src/main.rs
mkulke: Thanks, that's neat.

examples/rust/tracecon/src/bpf/tracecon.bpf.c
```diff
@@ -0,0 +1,168 @@
...
+
+struct event {
+	u8 tag;
+	u8 ip[4];
+	u8 hostname[HOSTNAME_LEN];
+};
+
+struct event _event = {0};
+
+static u32 get_tid() {
```
danobi, line 66: Just in case llvm decides to create a function one day

```suggestion
static __always_inline u32 get_tid() {
```

examples/rust/tracecon/src/bpf/tracecon.bpf.c
```diff
@@ -0,0 +1,168 @@
...
+	u64 tgid = bpf_get_current_pid_tgid();
+	pid_t pid = (pid_t)(tgid >> 32);
+	if (target_pid != 0 && target_pid != pid)
+		return 0;
+	return (u32)tgid;
+}
+
+SEC("uprobe/getaddrinfo")
+int BPF_KPROBE(getaddrinfo_enter,
```
danobi, line 75: Hmm, I suppose calling conventions should usually match and BPF_KPROBE() will be ok. Wonder why there isn't a BPF_UPROBE(). cc @anakryiko

examples/rust/tracecon/src/bpf/tracecon.bpf.c
```diff
@@ -0,0 +1,168 @@
...
+               const struct addrinfo *hints,
+               struct addrinfo **res)
+{
+	u32 tid = get_tid();
+	if (!tid)
+		return 0;
+	struct lookup lookup = {};
+	bpf_probe_read_user_str(&lookup.c, sizeof(lookup.c), hostname);
+	lookup.results = res;
+	bpf_map_update_elem(&lookups, &tid, &lookup, 0);
```
danobi, line 87: Better to use named flag
```suggestion
	bpf_map_update_elem(&lookups, &tid, &lookup, BPF_ANY);
```

examples/rust/tracecon/src/bpf/tracecon.bpf.c
```diff
@@ -0,0 +1,168 @@
...
+int BPF_KPROBE(getaddrinfo_enter,
+               const char *hostname,
+               const char *service,
+               const struct addrinfo *hints,
+               struct addrinfo **res)
+{
+	u32 tid = get_tid();
+	if (!tid)
+		return 0;
+	struct lookup lookup = {};
```
danobi, line 84: Put this at beginning of function (c89/kernel style)

examples/rust/tracecon/src/bpf/tracecon.bpf.c
```diff
@@ -0,0 +1,168 @@
...
+	return 0;
+}
+
+SEC("uretprobe/getaddrinfo")
+int BPF_KRETPROBE(getaddrinfo_exit, int ret)
+{
+	u32 tid = get_tid();
+	if (!tid)
+		return 0;
+	struct lookup *lookup = bpf_map_lookup_elem(&lookups, &tid);
```
danobi, line 97: Declare at beginning of function

examples/rust/tracecon/src/bpf/tracecon.bpf.c
```diff
@@ -0,0 +1,168 @@
...
+		struct sockaddr *aa;
+		struct in_addr ia;
+		u32 ip;
+		bpf_probe_read_user(&result, sizeof(result), lookup->results);
+		bpf_probe_read_user(&aa, sizeof(aa), &result->ai_addr);
+		bpf_probe_read_user(&ia, 
+		                    sizeof(ia),
+		                    &((struct sockaddr_in *) aa)->sin_addr);
+		ip = ia.s_addr;
+		bpf_map_update_elem(&hostnames, &ip, lookup, 0);
```
danobi, line 111:
```suggestion
		bpf_map_update_elem(&hostnames, &ip, lookup, BPF_ANY);
```

examples/rust/tracecon/src/bpf/tracecon.bpf.c
danobi: No, I think the map's value is `struct sock **`. See suggestion for what it should be instead.

examples/rust/tracecon/src/bpf/tracecon.bpf.c
```diff
@@ -0,0 +1,168 @@
...
+SEC("kprobe/tcp_v4_connect")
+int BPF_KPROBE(tcp_v4_connect_enter,
+               struct sock *sk,
+               struct sockaddr *uaddr,
+               int addr_len)
+{
+	u32 tid = get_tid();
+	if (!tid)
+		return 0;
+	bpf_map_update_elem(&sockets, &tid, &sk, 0);
```
danobi, line 126:
```suggestion
	bpf_map_update_elem(&sockets, &tid, sk, 0);
```

examples/rust/tracecon/src/bpf/tracecon.bpf.c
```diff
@@ -0,0 +1,168 @@
...
+int BPF_KRETPROBE(tcp_v4_connect_exit, int ret) {
+	u32 tid = get_tid();
+	if (!tid)
+		return 0;
+	struct sock **sockpp = bpf_map_lookup_elem(&sockets, &tid);
+	if (!sockpp)
+		return 0;
+  	if (ret == 0) {
+		struct event event = {};
+		struct sock *sockp = *sockpp;
```
danobi, line 140: Likewise this should be:
```suggestion
	struct sock *sockp = bpf_map_lookup_elem(&sockets, &tid);
	if (!sockp)
		return 0;
  	if (ret == 0) {
		struct event event = {};
```

examples/rust/tracecon/src/bpf/tracecon.bpf.c
```diff
@@ -0,0 +1,168 @@
...
+	u8 ip[4];
+	u8 hostname[HOSTNAME_LEN];
+};
+
+struct event _event = {0};
+
+static u32 get_tid() {
+	u64 tgid = bpf_get_current_pid_tgid();
+	pid_t pid = (pid_t)(tgid >> 32);
+	if (target_pid != 0 && target_pid != pid)
```
anakryiko, line 69: BPF-side  code generally follows kernel style, which is mostly old-fashioned C89, so variable declaration section has to be separated from other code with empty line

examples/rust/tracecon/src/bpf/tracecon.bpf.c
anakryiko: because uprobe and kprobe are exactly the same in kernel, so it felt redundant to add BPF_UPROBE/BPF_URETPROBE. I think ABI for passing arguments is the same as well, but I haven't checked. But basically, BPF_KPROBE() should be fine)

examples/rust/tracecon/src/bpf/tracecon.bpf.c
```diff
@@ -0,0 +1,168 @@
...
+		struct in_addr ia;
+		u32 ip;
+		bpf_probe_read_user(&result, sizeof(result), lookup->results);
+		bpf_probe_read_user(&aa, sizeof(aa), &result->ai_addr);
+		bpf_probe_read_user(&ia, 
+		                    sizeof(ia),
+		                    &((struct sockaddr_in *) aa)->sin_addr);
+		ip = ia.s_addr;
+		bpf_map_update_elem(&hostnames, &ip, lookup, 0);
+    }
```
anakryiko, line 112: indentation is off?

examples/rust/tracecon/src/bpf/tracecon.bpf.c
```diff
@@ -0,0 +1,168 @@
...
+	if (ret == 0) {
+		struct addrinfo *result;
+		struct sockaddr *aa;
+		struct in_addr ia;
+		u32 ip;
+		bpf_probe_read_user(&result, sizeof(result), lookup->results);
+		bpf_probe_read_user(&aa, sizeof(aa), &result->ai_addr);
+		bpf_probe_read_user(&ia, 
+		                    sizeof(ia),
+		                    &((struct sockaddr_in *) aa)->sin_addr);
```
anakryiko, line 109: this could be rewritten more nicely as 

```c
struct sockaddr_in *aa; /* define the actual type to avoid casts */

aa = BPF_PROBE_READ_USER(lookup, results, ai_addr);
ip = BPF_PROBE_READ_USER(aa, sin_addr.s_addr);
```

But you should really consider giving those variables some better names. And I'm not sure about double pointer in lookup...

examples/rust/tracecon/src/bpf/tracecon.bpf.c
anakryiko: @danobi, `sockets` map has `__type(value, struct sock *);`, so it seems fine? bpf_map_update_elem() expects a pointer to the value, so if the value is a pointer, you'll be passing a pointer to pointer. Everything checks out, unless I'm missing something.

examples/rust/tracecon/src/bpf/tracecon.bpf.c
anakryiko: you'll just read first 8 bytes of actual socket struct, which is probably not what you want

examples/rust/tracecon/src/bpf/tracecon.bpf.c
```diff
@@ -0,0 +1,168 @@
...
+
+SEC("kretprobe/tcp_v4_connect")
+int BPF_KRETPROBE(tcp_v4_connect_exit, int ret) {
+	u32 tid = get_tid();
+	if (!tid)
+		return 0;
+	struct sock **sockpp = bpf_map_lookup_elem(&sockets, &tid);
+	if (!sockpp)
+		return 0;
+  	if (ret == 0) {
```
anakryiko, line 138: I'd reduce nesting of the main logic and delete her and return early or `goto cleanup:` to avoid having two delete statements

examples/rust/tracecon/src/bpf/tracecon.bpf.c
```diff
@@ -0,0 +1,168 @@
...
+		struct event event = {};
+		struct sock *sockp = *sockpp;
+		struct sock_common sock;
+		bpf_probe_read_kernel(&sockp, sizeof(sockp), sockpp);
+		bpf_probe_read_kernel(&sock,
+		                      sizeof(sockp),
+		                      &sockp->__sk_common);
+		u32 daddr = sock.skc_daddr;
```
anakryiko, line 146: why reading sock_common if you need just one u32?

```c
daddr = BPF_CORE_READ(*sockpp, __sk_common.skc_daddr);
```

examples/rust/tracecon/src/bpf/tracecon.bpf.c
```diff
@@ -0,0 +1,168 @@
...
+		struct lookup *lookup = bpf_map_lookup_elem(&hostnames, &daddr);
+		if (!lookup) {
+			event.tag = IP;
+			memcpy(&event.ip, &daddr, sizeof(event.ip));
```
anakryiko, line 150: there is no memcpy() on BPF side, you are just getting like for compiler to inline it for you. Use `__builtin_memcpy()` instead

anakryiko: but actually is that such a big deal to have a properly aligned u32 in your event struct?

examples/rust/tracecon/src/bpf/tracecon.bpf.c
```diff
@@ -0,0 +1,168 @@
...
+		}
+		else {
```
anakryiko, line 152: } else {

examples/rust/tracecon/src/bpf/tracecon.bpf.c
```diff
@@ -0,0 +1,168 @@
...
+		                      &events,
+		                      BPF_F_CURRENT_CPU,
+		                      &event,
+		                      sizeof(event));
+  	}
+ 	bpf_map_delete_elem(&sockets, &tid);
+	return 0;
+}
+
+char LICENSE[] SEC("license") = "Dual BSD/GPL";
```
anakryiko, line 168: can you please also add SPDX header(s) at the top of file(s)

examples/rust/tracecon/src/bpf/tracecon.bpf.c
mkulke: thx, that's a good use case for the BPF macros 👍

examples/rust/tracecon/src/bpf/tracecon.bpf.c
mkulke: I used __builtin it's aliased w/ a #define on the top. The u32 should be aligned, otherwise we'd have to do it on the rust client code.

examples/rust/tracecon/README.md
```diff
@@ -0,0 +1,76 @@
...
+
+```bash
+cat /boot/config-$(uname -a) | grep CONFIG_DEBUG_INFO_BTF
+```
+
+## Build
+
+### Vagrant
+
+eBPF is a low-level technology on the Linux kernel. Docker is not a good fit to build eBPF code on MacOS or Windows environments. On those platforms Docker ships its own kernel (e.g. linuxkit) and the kernel headers of the image might not match the running kernel.
```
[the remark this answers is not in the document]
mkulke: Right, that's inaccurate, fixed the remark. What's actually missing is BTF support on linuxkit kernels in Docker's VM:

```
bpftrace --info 2> >(grep -C3 btf)
  bpf_attach_kfunc: no
  bcc_usdt_addsem: no
  libbpf: no
  libbpf btf dump: no
  libbpf btf dump type decl: no

Kernel helpers
  probe_read: yes
--
Kernel features
  Instruction limit: 1000000
  Loop support: yes
  btf: no

Map types
  hash: yes
```

hence:

```
docker run --privileged -it mgns /tracecon
libbpf: failed to find valid kernel BTF
libbpf: Error loading vmlinux BTF: -3
libbpf: failed to load object 'tracecon_bpf'
libbpf: failed to load BPF skeleton 'tracecon_bpf': -3
Error: System error, errno: 3
```

mkulke:
@danobi @anakryiko Thank you for your thorough review, that helped a lot. I hope I was able to address all your remarks, PTAL

(There is one remaining conversation about passing ptr to ptr in the call to populate a map. but as @anakryiko remarked, i think it has to be like this)

examples/rust/tracecon/src/bpf/tracecon.bpf.c
anakryiko: you need to have an empty line between variable declarations and the rest of the code. Here, above this line, and in few more places around the code

examples/rust/tracecon/src/bpf/tracecon.bpf.c
```diff
@@ -0,0 +1,161 @@
...
+	struct addrinfo *result;
+	struct sockaddr *addr;
+	struct in_addr ipv4_addr;
+	if (!tid)
+		return 0;
+	lookup = bpf_map_lookup_elem(&lookups, &tid);
+	if (!lookup)
+		return 0;
+	if (ret != 0)
+		goto cleanup;
```
anakryiko, line 103: you should check this before doing lookup, why doing unnecessary work?

examples/rust/tracecon/src/bpf/tracecon.bpf.c
```diff
@@ -0,0 +1,161 @@
...
+	bpf_probe_read_user(&result, sizeof(result), lookup->results);
+	bpf_probe_read_user(&addr, sizeof(addr), &result->ai_addr);
+	bpf_probe_read_user(&ipv4_addr, sizeof(ipv4_addr),
+			    &((struct sockaddr_in *)addr)->sin_addr);
```
anakryiko, line 107: Have you tried BPF_PROBE_READ_USER() macro for this?

examples/rust/tracecon/src/bpf/tracecon.bpf.c
```diff
@@ -0,0 +1,161 @@
...
+	bpf_map_update_elem(&hostnames, &ipv4_addr.s_addr, lookup, BPF_ANY);
+cleanup:
+	bpf_map_delete_elem(&lookups, &tid);
+	return 0;
+}
+
+SEC("kprobe/tcp_v4_connect")
+int BPF_KPROBE(tcp_v4_connect_enter, struct sock *sk, struct sockaddr *uaddr,
+	       int addr_len)
```
anakryiko, line 116: nit: keep on single line, not worth to wrap

examples/rust/tracecon/src/bpf/tracecon.bpf.c
```diff
@@ -0,0 +1,161 @@
...
+	u32 ip;
+
+	if (!tid)
+		return 0;
+
+	sockpp = bpf_map_lookup_elem(&sockets, &tid);
+	if (!sockpp)
+		return 0;
+	if (ret != 0)
+		goto cleanup;
```
anakryiko, line 141: same, check ret first, then do lookup

examples/rust/tracecon/src/bpf/tracecon.bpf.c
```diff
@@ -0,0 +1,161 @@
...
+		event.tag = IP;
+		memcpy(&event.ip, &ip, sizeof(event.ip));
+	} else {
+		event.tag = HOSTNAME;
+		memcpy(&event.hostname, &lookup->c, sizeof(lookup->c));
+		bpf_map_delete_elem(&hostnames, &ip);
+	}
+	/* ctx is implied in the signature macro */
+	bpf_perf_event_output(ctx, &events, BPF_F_CURRENT_CPU, &event,
+			      sizeof(event));
```
anakryiko, line 155: nit: I'd keep single line

anakryiko (changes requested):
Please squash all those 19 commits into one. It would also be good to have a bit better description in the commit message with an example of expected output.

We should probably have some mention in main README.md at the root of the repo listing all the advanced examples, so that people are aware of what's there (maybe with a single sentence description of what each example demonstrates). WDYT?

examples/rust/tracecon/src/bpf/tracecon.bpf.c
mkulke: The macro is not available in the version included in Ubuntu LTS 20.10, I installed libbpf v0.4 and tried it out, due to the cast it's not really less verbose, and maybe a bit more obscure:

```c
        bpf_probe_read_user(&result, sizeof(result), lookup->results);
        addr = BPF_PROBE_READ_USER(result, ai_addr);
        ipv4_addr = BPF_PROBE_READ_USER((struct sockaddr_in *) addr, sin_addr);
```

i would opt for leaving the non-macro calls, wdyt?

examples/rust/Cargo.toml
```diff
@@ -1,4 +1,5 @@
 [workspace]
 members = [
   "xdp",
+  "tracecon",
```
anakryiko, line 4: if this file is not auto-generated, let's keep projects alphabetically sorted

examples/rust/tracecon/src/bpf/tracecon.bpf.c
```diff
@@ -0,0 +1,162 @@
...
+		return 0;
+	if (ret != 0)
+		goto cleanup;
+	lookup = bpf_map_lookup_elem(&lookups, &tid);
+	if (!lookup)
+		return 0;
+	bpf_probe_read_user(&result, sizeof(result), lookup->results);
+	bpf_probe_read_user(&addr, sizeof(addr), &result->ai_addr);
+	bpf_probe_read_user(&ipv4_addr, sizeof(ipv4_addr),
+	                    &((struct sockaddr_in *)addr)->sin_addr);
```
anakryiko, line 110: Ok, sure, you can also declare `struct sockaddr_in *addr` instead of `struct sockaddr *addr;`, then you'll avoid extra cast.

examples/rust/tracecon/src/bpf/tracecon.bpf.c
danobi: Ah damn you're right. I don't think this is the first time I've messed that up either.

examples/rust/tracecon/src/bpf/tracecon.bpf.c
```diff
@@ -0,0 +1,162 @@
...
+
+enum tag { IP = 0, HOSTNAME = 1 };
+
+struct event {
+	u8 tag;
+	u8 ip[4];
+	u8 hostname[HOSTNAME_LEN];
+};
+
+struct event _event = {};
```
danobi, line 64: Nit: maybe leave a comment why this exists

danobi (approved):
LGTM, just a few outstanding comments left I think

anakryiko (approved):
LGTM. It would be good to follow up with mentions of tracecon in main README.md so that people can discover it more easily.

PR #33, opened 2021-06-14
Remove Rust's redundant build commands

tallossos:
Since the example uses a build script, using the cargo plugin commands in the example is redundant.
Removing it would clarify the build process.


danobi (approved):
Ah whoops. Thanks!

PR #34, opened 2021-06-21
examples/c: fix build failure(llvm-strip not found)

yangxingwu:


examples/c/Makefile
```diff
@@ -1,7 +1,7 @@
 # SPDX-License-Identifier: (LGPL-2.1 OR BSD-2-Clause)
 OUTPUT := .output
 CLANG ?= clang
-LLVM_STRIP ?= llvm-strip
+LLVM_STRIP ?= $(shell dirname `readlink -f \`which $(CLANG)\``)/llvm-strip
```
anakryiko, line 4: is the idea here that CLANG will be overriden with the path to clang binary and we want to use the same directory to find llvm-strip? If yes, then why user can't override LLVM_STRIP directly? If not, why llvm-strip was not installed along the clang (but apparently is there along the clang binary).

It would be nice to have a bit more context in your PR explaining what motivated you to create this PR. Thanks

yangxingwu: the idea is clang is in the `$PATH` but llvm-strip not

```bash
$ which clang
/usr/bin/clang

$ llvm-strip
command not found: llvm-strip

$ which llvm-strip-12
/usr/bin/llvm-strip-12
```

build env:

```bash
$ lsb_release -a
No LSB modules are available.
Distributor ID:	Ubuntu
Description:	Ubuntu 21.04
Release:	21.04
Codename:	hirsute
```

anakryiko: This sounds like either the problem with llvm package on Ubuntu or you are missing some extra package (e.g., something like llvm-tools, don't know, not an Ubuntu user). Assuming that llvm-strip is in the same directory as clang isn't the right assumption to make, in general, so I'm not inclined to make this change. Worst case, you can always override LLVM_STRIP variable pointing to the exact llvm-strip binary you need.

yangxingwu: I test again and find that package llvm provides llvm-strip, so problem can be resolved by install llvm package

I will close this PR soon

```bash
$ sudo apt-file find llvm-strip
llvm: /usr/bin/llvm-strip
llvm-11: /usr/bin/llvm-strip-11
llvm-11: /usr/lib/llvm-11/bin/llvm-strip
llvm-11-doc: /usr/share/doc/llvm-11-doc/html/CommandGuide/llvm-strip.html
llvm-11-doc: /usr/share/doc/llvm-11-doc/html/_sources/CommandGuide/llvm-strip.rst.txt
llvm-12: /usr/bin/llvm-strip-12
llvm-12: /usr/lib/llvm-12/bin/llvm-strip
llvm-12-doc: /usr/share/doc/llvm-12-doc/html/CommandGuide/llvm-strip.html
llvm-12-doc: /usr/share/doc/llvm-12-doc/html/_sources/CommandGuide/llvm-strip.rst.txt
llvm-9: /usr/bin/llvm-strip-9
llvm-9: /usr/lib/llvm-9/bin/llvm-strip
llvm-9-doc: /usr/share/doc/llvm-9-doc/html/CommandGuide/llvm-strip.html
llvm-9-doc: /usr/share/doc/llvm-9-doc/html/_sources/CommandGuide/llvm-strip.md.txt
```

PR #51, opened 2021-11-16
fix missing `-g` flag in xmake.lua

waruqi:


examples/c/xmake.lua
```diff
@@ -14,6 +14,7 @@ else
     set_toolchains("@llvm")
     add_requires("linux-headers")
 end
+set_symbols("debug")
```
anakryiko, line 17: will this apply to only .bpf.c? I assume not. Is there a way to control this more precisely?

I'll merge as for bootstrap examples having debug info everywhere is nice anyways, but -g is *required* only for *.bpf.c files.

PR #52, opened 2021-11-17
Update xmake.lua

waruqi:


anakryiko:
I've updated the error message and force-pushed. Please check if it makes sense. I've also dropped the -g commit.

waruqi:
thanks!

Issue #70, opened 2022-05-09
Building using cmake is not working


mariuszbarczak:
When trying to build the project using cmake, I've got the following error:

bash: line 1: /home/mariusz/git/libbpf-bootstrap/examples/c/../../tools/bpftool: No such file or directory
make[2]: *** [CMakeFiles/bootstrap.dir/build.make:74: bootstrap.skel.h] Error 127
make[2]: *** Deleting file 'bootstrap.skel.h'
make[1]: *** [CMakeFiles/Makefile2:145: CMakeFiles/bootstrap.dir/all] Error 2
make: *** [Makefile:91: all] Error 2

mariuszbarczak:
I added ExternalProject_Add in examples/c/CMakeLists.txt for bpftool, it started working. Is it acceptable?

anakryiko:
@danobi can you please take a look if you get a chance? But either way, @JackyYin, please go ahead and submit your fix, thank you!

danobi:
The submitted fix lgtm

anakryiko:
Closed by https://github.com/libbpf/libbpf-bootstrap/pull/73

mariuszbarczak:
Thanks Guys for fixing this.

PR #73, opened 2022-05-19
examples/c: Build bpftool submodule in cmake

JackyYin:
Ref: #70.

Because tools/bpftool no longer exists, we need to build bpftool from git submodule and set the correct executable path in cmake variable `BPFOBJECT_BPFTOOL_EXE`.

### DIFF
--- a/examples/c/CMakeLists.txt
+++ b/examples/c/CMakeLists.txt
@@ -25,9 +25,20 @@ ExternalProject_Add(libbpf
   STEP_TARGETS build
 )
 
+ExternalProject_Add(bpftool
+  PREFIX bpftool
+  SOURCE_DIR ${CMAKE_CURRENT_SOURCE_DIR}/../../bpftool/src
+  CONFIGURE_COMMAND ""
+  BUILD_COMMAND make
+    OUTPUT=${CMAKE_CURRENT_BINARY_DIR}/bpftool/
+  BUILD_IN_SOURCE TRUE
+  INSTALL_COMMAND ""
+  STEP_TARGETS build
+)
+
 # Set BpfObject input parameters -- note this is usually not necessary unless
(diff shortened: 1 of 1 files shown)


JackyYin:
BTW, seems like we can remove the `STEP_TARGETS` in `ExternalProject_Add` and this line also: `add_dependencies(${app_stem}_skel libbpf-build)` because libbpf always built before examples even if I remove these lines.

JackyYin:
Hello @anakryiko,
Could you give some suggestions about this PR?
I wonder what can I fix to get it merged.
Thanks a lot!

examples/c/CMakeLists.txt
```diff
@@ -25,9 +25,20 @@ ExternalProject_Add(libbpf
   STEP_TARGETS build
 )
 
+ExternalProject_Add(boftool
```
anakryiko, line 28: this should probably be `bpftool`, no?

anakryiko (changes requested):
I see this follows the approach for building libbpf so looks sane to me. Please fix the typo and can you also add Signed-off-by: with your full name and email just like libbpf/kernel commits have it? Thanks!

examples/c/CMakeLists.txt
JackyYin: Fixed.
Sorry for the typo.

JackyYin:
> I see this follows the approach for building libbpf so looks sane to me. Please fix the typo and can you also add Signed-off-by: with your full name and email just like libbpf/kernel commits have it? Thanks!

Thanks for your feedback!
I just fixed the typo and added the `Signed-off-by` to my commit message.
The commit is squashed and force-pushed back again.

anakryiko (approved):

anakryiko:
Thanks!

danobi:
lgtm

PR #78, opened 2022-05-25
Add examples of blazesym

ThinkerYzu1:
Include two examples, bzsstack for Rust and profile for C.
bzsstack dumps a stack trace on the raw tracepoint of sys_enter.
profile is a profiler to sample stack traces of kernel and userspace every second.
Both of them use blazesym to symbolize addresses to get symbol names, file names, and line numbers.


examples/c/profile.c
```diff
@@ -0,0 +1,209 @@
...
+int
+main(int argc, char * const argv[])
+{
+	int freq = 1, pid = -1, cpu;
+	struct profile_bpf *skel = NULL;
+	struct perf_event_attr attr;
+	struct bpf_link **links = NULL;
+	struct handler_ctx ctx;
+	struct ring_buffer *ring_buf = NULL;
+	int num_cpus = get_nprocs(); /* number of processors */
```
anakryiko, line 116: use libbpf_num_possible_cpus(), there is some long history with get_nprocs() not working for all possible use cases properly, stick to what libbpf provides

examples/c/profile.c
```diff
@@ -0,0 +1,209 @@
...
+				    NULL);
+	if (ring_buf == NULL) {
+		err = -1;
+		goto cleanup;
+	}
+
+	pefds = (int *)malloc(sizeof(int) * num_cpus);
+	memset(pefds, -1, sizeof(int) * num_cpus);
+
+	links = (struct bpf_link **)malloc(sizeof(struct bpf_link *) * num_cpus);
```
anakryiko, line 158: use calloc() to not have to memset(0)

examples/c/profile.c
```diff
@@ -0,0 +1,209 @@
...
+	memset(links, 0, sizeof(struct bpf_link *) * num_cpus);
+
+	memset(&attr, 0, sizeof(attr));
+	attr.type = PERF_TYPE_HARDWARE;
+	attr.size = sizeof(attr);
+	attr.config = PERF_COUNT_HW_CPU_CYCLES;
+	attr.sample_freq = freq;
```
anakryiko, line 165: it would be good for freq to be an input argument that user can specify

examples/c/profile.c
```diff
@@ -0,0 +1,209 @@
...
+	attr.freq = 1;
+
+	for (cpu = 0; cpu < num_cpus; cpu++) {
+		/* Set up performance monitoring on a CPU/Core */
+		pefd = perf_event_open(&attr, pid, cpu, -1, PERF_FLAG_FD_CLOEXEC);
+		if (pefd < 0) {
+			perror("perf_event_open");
```
anakryiko, line 172: stay consistent with how you report errors, you've been using fprintf(stderr) up till this point

examples/c/profile.c
```diff
@@ -0,0 +1,209 @@
...
+	}
+
+	/* Wait and receive stack traces */
+	while (ring_buffer__poll(ring_buf, -1) >= 0) {
+	}
+
+cleanup:
+	if (links) {
+		for (cpu = 0; cpu < num_cpus; cpu++)
+			if (links[cpu])
```
anakryiko, line 192: bpf_link__destroy() handles NULL properly, so you don't need this extra if check

examples/c/profile.c
```diff
@@ -0,0 +1,209 @@
...
+				bpf_link__destroy(links[cpu]);
+		free(links);
+	}
+	if (pefds) {
+		for (i = 0; i < num_cpus; i++)
+			if (pefds[i] >= 0)
+				close(pefds[i]);
+		free(pefds);
+	}
+	if (ring_buf)
```
anakryiko, line 202: same, no need to check, all libbpf "destructor" APIs handle NULL properly

examples/c/profile.c
```diff
@@ -0,0 +1,209 @@
...
+		ring_buffer__free(ring_buf);
+	if (skel)
```
anakryiko, line 204: ditto

examples/c/profile.c
```diff
@@ -0,0 +1,209 @@
...
+		profile_bpf__destroy(skel);
+	if (ctx.symbolizer)
```
anakryiko, line 206: and please make sure that blazesym library handles NULL properly as well, it's a nice touch for an API (libc's free() does that as well, btw)

examples/c/profile.h
```diff
@@ -0,0 +1,20 @@
...
+struct stacktrace_event {
+	__u32 pid;
+	__u32 cpu_id;
+	__s32 kernel_stackid;
+	__s32 user_stackid;
+};
+
+typedef __u64 stack_trace_t[MAX_STACK_DEPTH];
+
+
```
anakryiko, line 19: extra line

examples/c/profile.h
```diff
@@ -0,0 +1,20 @@
+/* SPDX-License-Identifier: (LGPL-2.1 OR BSD-2-Clause) */
+/* Copyright (c) 2022 Facebook */
```
anakryiko, line 2: s/Facebook/Meta Platforms, Inc./

examples/rust/bzsstack/src/main.rs
```diff
@@ -0,0 +1,95 @@
...
+
+    let skel_builder = BzsstackSkelBuilder::default();
+
+    let mut open_skel = skel_builder.open().unwrap();
+    let pid = getpid().as_raw();
+    open_skel.bss().my_pid = pid;
+    let mut skel = open_skel.load().unwrap();
+
+    let _handle_tp =
+	skel.progs_mut().handle_tp().attach().unwrap();
```
anakryiko, line 41: is this how cargo fmt format this? If you haven't please run `cargo fmt` so that everything is formatted consistently

examples/c/profile.c
```diff
@@ -0,0 +1,209 @@
...
+		int cpu, int group_fd, unsigned long flags)
+{
+	int ret;
+
+	ret = syscall(__NR_perf_event_open, hw_event, pid, cpu,
+		      group_fd, flags);
+	return ret;
+}
+
+struct handler_ctx {
```
anakryiko, line 28: for a simple tool like this it is ok to use static global variables to keep this process-wide state. We do that a lot in libbpf-tools and it works totally fine. The only downside to that is worse testability through unit tests, but no one is going to write unit tests for these demos anyways, so there is no need to add extra verboseness for this. Keep it simple.

examples/c/profile.c
```diff
@@ -0,0 +1,209 @@
...
+{
+	const struct blazesym_result *syms;
+	int cnt, i;
+
+	for (cnt = 0; stack[cnt]; cnt++) {
+	}
+	syms = blazesym_symbolize(ctx->symbolizer, ctx->cfgs, 2, (const uint64_t *)stack, cnt);
+	for (i = 0; stack[i]; i++) {
+		if (syms[i].valid)
+			if (syms[i].path && syms[i].path[0])
```
anakryiko, line 45: it's my personal preference, but even though printf() below is a single statement and kernel code style allows to not have {} for such if/for/while statements, I think it looks bad if the body of if/for/while is more than a single line, so can you please add {} around multi-line statements? Thanks!

ThinkerYzu1:
Got it!

examples/c/Makefile
```diff
@@ -64,6 +75,14 @@ $(BPFTOOL): | $(dir $(BPFTOOL))
 	$(call msg,BPFTOOL,$@)
 	$(Q)$(MAKE) OUTPUT=$(dir $(BPFTOOL)) -C $(BPFTOOL_SRC)
 
+$(LIBBLAZESYM_OBJ):: $(LIBBLAZESYM_SRC)/src/*.rs $(LIBBLAZESYM_SRC)/Cargo.toml $(OUTPUT)
```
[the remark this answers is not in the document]
ThinkerYzu1: I should remove *.rs but keep double colons.  The directory of blazesym is getting complicated (submodules).  It is unrealistic to maintain dependencies here.  I want to run cargo to check dependencies for me.

Does that make sense?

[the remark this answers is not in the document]
ThinkerYzu1: I found it does not work as I expected.  I will change it.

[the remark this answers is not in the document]
anakryiko: oh, right, forgot to ask if it's possible to basically always run cargo instead of guessing dependencies. Ideally cargo will know when nothing should be done and skip building stuff, right? So basically you add FORCE as depenency to always call cargo build. You'll probably need to be careful to not copy .a file if it didn't change (or at least preserve original modification time), otherwise we'll keep rebuilding profile app all the time, even if nothing changed.

examples/c/Makefile
```diff
@@ -14,7 +17,15 @@ VMLINUX := ../../vmlinux/$(ARCH)/vmlinux.h
...
-APPS = minimal bootstrap uprobe kprobe fentry
+CARGO ?= $(shell which cargo)
+ifeq ($(strip $(CARGO)),)
+BZS_APPS :=
+else
+BZS_APPS := profile
+# Required by libblazesym
+LDFLAGS-y += -lrt -ldl -lpthread
+endif
+APPS = minimal bootstrap uprobe kprobe fentry $(BZS_APPS)
```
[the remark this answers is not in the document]
ThinkerYzu1: Make sense

examples/rust/bzsstack/src/main.rs
ThinkerYzu1: Roger1

examples/c/profile.c
ThinkerYzu1: Sure!  I personally prefer always to have braces.

ThinkerYzu1:
Just update the PR to make the rust example have the same features as the C example.

Also, make sampling frequency a command-line option.

examples/c/profile.bpf.c
```diff
@@ -0,0 +1,40 @@
...
+#include <bpf/bpf_tracing.h>
+#include <bpf/bpf_core_read.h>
+
+#include "profile.h"
+
+char LICENSE[] SEC("license") = "Dual BSD/GPL";
+
+struct {
+	__uint(type, BPF_MAP_TYPE_RINGBUF);
+	__uint(max_entries, sizeof(struct stacktrace_event) * 64);
```
anakryiko, line 14: with most recent libbpf this would work because libbpf will auto-adjust the size to the closest multiple of page size, but I think it's a bit confusing to express the size like this. Let's just have something like `256 * 1024` specified as a size?

examples/c/profile.bpf.c
```diff
@@ -0,0 +1,40 @@
...
+	struct stacktrace_event *event =
+		bpf_ringbuf_reserve(&events, sizeof(struct stacktrace_event), 0);
+
+	if (!event)
+		return 1;
+	event->pid = pid;
+	event->cpu_id = cpu_id;
+
+	event->kernel_stack_sz = bpf_get_stack(ctx, event->kernel_stack,
+					       MAX_STACK_DEPTH * sizeof(event->kernel_stack[0]),
```
anakryiko, line 31: few problems here:

1. you are losing error code if error happens if you divide by sizeof(), do that division in user-space instead
2. use `sizeof(event->kernel_stack)` and don't duplicate MAX_STACK_DEPTH here. Again, if you change something about kernel_stack array you won't have to change this part.
3. Also just making sure (in case it's not Github UI formatting issue) that when wrapping function call, first wrapped argument on each line aligns with first argument on previous line.

examples/c/profile.c
```diff
@@ -0,0 +1,197 @@
...
+#include <linux/perf_event.h>
+#include <bpf/libbpf.h>
+#include <bpf/bpf.h>
+
+#include "profile.skel.h"
+#include "profile.h"
+#include "blazesym.h"
+
+static long perf_event_open(struct perf_event_attr *hw_event, pid_t pid,
+		int cpu, int group_fd, unsigned long flags)
```
anakryiko, line 17: see above about function argument alignment (if it's not Github screwing tabs, of course)

examples/c/profile.c
```diff
@@ -0,0 +1,197 @@
...
+				       i, stack[i], syms[i].symbol,
+				       stack[i] - syms[i].start_address,
+				       syms[i].path, syms[i].line_no);
+			} else {
+				printf("  %d [<%016lx>] %s+0x%lx\n",
+				       i, stack[i], syms[i].symbol,
+				       stack[i] - syms[i].start_address);
+			}
+		} else
+			printf("  %d [<%016lx>]\n", i, stack[i]);
```
anakryiko, line 62: if one part of if/else has {} then the other has to as well (even if it's a simple single-line statement)

examples/c/profile.c
```diff
@@ -0,0 +1,197 @@
...
+	}
+	if (syms)
+		blazesym_result_free(syms);
+}
+
+static int event_handler(void *_ctx, void *data, size_t size)
+{
+	struct stacktrace_event *event = (struct stacktrace_event *)data;
```
anakryiko, line 70: in C you don't have to explicitly cast void pointer, so you can drop this cast completely

examples/c/profile.c
```diff
@@ -0,0 +1,197 @@
...
+
+	if (event->kernel_stack_sz == 0 && event->user_stack_sz == 0)
```
anakryiko, line 72: on error you'll have negative values for these sizes, not zero

examples/c/profile.c
```diff
@@ -0,0 +1,197 @@
...
+		return 1;
+
+	printf("PID %d, CPU %d\n", event->pid, event->cpu_id);
+
+	if (event->kernel_stack_sz > 0) {
+		printf("Kernel:\n");
+		show_stack_trace((uint64_t *)event->kernel_stack, event->kernel_stack_sz, 0);
```
anakryiko, line 79: is this cast necessary?

examples/c/profile.c
```diff
@@ -0,0 +1,197 @@
...
+	} else
+		printf("No Kernel Stack\n");
```
anakryiko, line 81: missing {}

examples/c/profile.c
```diff
@@ -0,0 +1,197 @@
...
+
+	if (event->user_stack_sz > 0) {
+		printf("Userspace:\n");
+		show_stack_trace((uint64_t *)event->user_stack, event->user_stack_sz, event->pid);
+	} else
+		printf("No Userspace Stack\n");
```
anakryiko, line 87: ditto, {}

examples/c/profile.c
```diff
@@ -0,0 +1,197 @@
...
+	if (pid) {
+		cfg.cfg_type = CFG_T_PROCESS;
+		cfg.params.process.pid = pid;
+	} else {
+		cfg.cfg_type = CFG_T_KERNEL;
+		cfg.params.kernel.kallsyms = NULL;
+		cfg.params.kernel.kernel_image = NULL;
+	}
+
+	syms = blazesym_symbolize(symbolizer, &cfg, 1, (const uint64_t *)stack, stack_sz);
```
anakryiko, line 48: unnecessary cast?

examples/c/profile.c
```diff
@@ -0,0 +1,197 @@
...
+	for (i = 0; i < stack_sz; i++) {
+		if (syms && syms[i].valid) {
+			if (syms[i].path && syms[i].path[0]) {
+				printf("  %d [<%016lx>] %s+0x%lx %s:%ld\n",
```
anakryiko, line 52: hm... ok, what am I missing? Single memory address can resolve to multiple symbolized entries due to inlined functions. But here you have a loop assuming one-to-one mapping. How would inline function be communicated? And how I would request inline function to be resolved using blazesym API?

examples/c/profile.c
```diff
@@ -0,0 +1,197 @@
...
+	int freq = 1, pid = -1, cpu;
+	struct profile_bpf *skel = NULL;
+	struct perf_event_attr attr;
+	struct bpf_link **links = NULL;
+	struct ring_buffer *ring_buf = NULL;
+	int num_cpus = libbpf_num_possible_cpus(); /* number of processors */
+	int *pefds = NULL, pefd;
+	int opt, i, err;
+
+	while ((opt = getopt(argc, argv, "hf:p:d:")) != -1) {
```
anakryiko, line 109: let's use argp instead, same as bootstrap.c example

examples/c/profile.c
```diff
@@ -0,0 +1,197 @@
...
+
+	skel = profile_bpf__open_and_load();
+	if (!skel) {
+		fprintf(stderr, "Failed to open and load BPF skeleton\n");
+		return 1;
+	}
+
+	/* Data passing to event_handler */
+	symbolizer = blazesym_new();
+	if (symbolizer == NULL) {
```
anakryiko, line 132: nit: `if (!symbolizer)`

examples/c/profile.c
```diff
@@ -0,0 +1,197 @@
...
+		fprintf(stderr, "Fail to create a symbolizer\n");
+		err = -1;
+		goto cleanup;
+	}
+
+	/* Prepare ring buffer */
+	ring_buf = ring_buffer__new(bpf_map__fd(skel->maps.events),
```
anakryiko, line 139: nit: keep it on a single line, it's not that long?

examples/c/profile.c
```diff
@@ -0,0 +1,197 @@
...
+				    event_handler,
+				    NULL,
+				    NULL);
+	if (ring_buf == NULL) {
```
anakryiko, line 143: it's unusual for C code to do `== NULL`, pointers are usually checked as `!ptr`

examples/c/profile.c
```diff
@@ -0,0 +1,197 @@
...
+		err = -1;
+		goto cleanup;
+	}
+
+	pefds = (int *)calloc(num_cpus, sizeof(int));
```
anakryiko, line 148: no need to cast

examples/c/profile.c
```diff
@@ -0,0 +1,197 @@
...
+
+	links = (struct bpf_link **)calloc(num_cpus, sizeof(struct bpf_link *));
```
anakryiko, line 150: same, no casting. And see my previous suggestions, stick to `sizeof(*links)`

examples/c/profile.c
```diff
@@ -0,0 +1,197 @@
...
+		pefd = perf_event_open(&attr, pid, cpu, -1, PERF_FLAG_FD_CLOEXEC);
+		if (pefd < 0) {
+			fprintf(stderr, "Fail to set up performance monitor on a CPU/Core\n");
+			goto cleanup;
+		}
+		pefds[cpu] = pefd;
+
+		/* Attach a BPF program on a CPU */
+		links[cpu] = bpf_program__attach_perf_event(skel->progs.profile, pefd);
+		if (links[cpu] == NULL) {
```
anakryiko, line 170: !links[cpu]

examples/c/profile.h
anakryiko: ?

examples/rust/Cargo.toml
```diff
@@ -2,4 +2,5 @@
 members = [
   "tracecon",
   "xdp",
+  "bzsstack",
```
anakryiko, line 5: it's hard to pronounce it, but also why using a different name from C example? Just "profile"?

examples/rust/bzsstack/src/bpf/profile.bpf.c
```diff
@@ -0,0 +1,40 @@
+// SPDX-License-Identifier: GPL-2.0 OR BSD-3-Clause
```
anakryiko, line 1: let's just symlink this to profile.bpf.c from C example, why maintaining two copies?

examples/rust/bzsstack/src/main.rs
```diff
@@ -0,0 +1,201 @@
...
+            kallsyms: None,
+            kernel_image: None,
+        }
+    } else {
+        SymbolFileCfg::Process { pid: Some(pid) }
+    };
+
+    let syms = symbolizer.symbolize(&[cfg], stack);
+    for i in 0..stack.len() {
+        if syms.len() == stack.len() {
```
anakryiko, line 88: see my comments for C example, this doesn't seem right

examples/rust/bzsstack/src/main.rs
```diff
@@ -0,0 +1,201 @@
...
+
+    let skel_builder = ProfileSkelBuilder::default();
+    let open_skel = skel_builder.open().unwrap();
+    let mut skel = open_skel.load().unwrap();
+
+    let pefds = init_perf_monitor(freq);
+    let _links = attach_perf_event(&pefds, skel.progs_mut().profile());
+
+    let ring_buf = unsafe {
+        libbpf_sys::ring_buffer__new(
```
anakryiko, line 186: libbpf-rs has Rust bindings for ringbuf, you shouldn't need to have so much unsafe code.

@insearchoflosttime, can you please help reviewing Rust parts and suggesting best libbpf-rs usage? Thanks!

examples/rust/bzsstack/src/syscall.rs
```diff
@@ -0,0 +1,81 @@
...
+#[repr(C)]
+pub union bp_2_un {
+    pub bp_len: u64,
+    pub kprobe_addr: u64,
+    pub probe_offset: u64,
+    pub config2: u64,
+}
+
+#[repr(C)]
+pub struct perf_event_attr {
```
anakryiko, line 32: btw, latest libbpf-sys package (which libbpf-rs is also using) should have all the perf_event definitions, so you should be able to reuse it

examples/c/profile.bpf.c
ThinkerYzu1: Ok!

examples/c/profile.bpf.c
```diff
@@ -0,0 +1,40 @@
...
+	__uint(max_entries, sizeof(struct stacktrace_event) * 64);
+} events SEC(".maps");
+
+SEC("perf_event")
+int profile(void *ctx)
+{
+	int pid = bpf_get_current_pid_tgid() >> 32;
+	int cpu_id = bpf_get_smp_processor_id();
+	struct stacktrace_event *event =
+		bpf_ringbuf_reserve(&events, sizeof(struct stacktrace_event), 0);
```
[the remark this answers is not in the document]
ThinkerYzu1: I will rename user_stack and kernel_stack to ustck and kstack .

examples/c/profile.c
```diff
@@ -0,0 +1,197 @@
...
+#include "profile.skel.h"
+#include "profile.h"
+#include "blazesym.h"
+
+static long perf_event_open(struct perf_event_attr *hw_event, pid_t pid,
+		int cpu, int group_fd, unsigned long flags)
+{
+	int ret;
+
+	ret = syscall(__NR_perf_event_open, hw_event, pid, cpu,
```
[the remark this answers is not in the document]
ThinkerYzu1: Roger

examples/c/profile.c
ThinkerYzu1: Got it

examples/c/profile.c
```diff
@@ -0,0 +1,197 @@
...
+			} else {
+				printf("  %d [<%016lx>] %s+0x%lx\n",
+				       i, stack[i], syms[i].symbol,
+				       stack[i] - syms[i].start_address);
+			}
+		} else
+			printf("  %d [<%016lx>]\n", i, stack[i]);
+	}
+	if (syms)
+		blazesym_result_free(syms);
```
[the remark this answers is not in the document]
ThinkerYzu1: You are right!

examples/c/profile.c
ThinkerYzu1: I think it would be easier to set them to 0 if bpf_get_stack() fails.

examples/c/profile.c
ThinkerYzu1: Same as I mentioned earlier.

examples/c/profile.c
```diff
@@ -0,0 +1,197 @@
...
+	pefds = (int *)calloc(num_cpus, sizeof(int));
+
+	links = (struct bpf_link **)calloc(num_cpus, sizeof(struct bpf_link *));
+
+	memset(&attr, 0, sizeof(attr));
+	attr.type = PERF_TYPE_HARDWARE;
+	attr.size = sizeof(attr);
+	attr.config = PERF_COUNT_HW_CPU_CYCLES;
+	attr.sample_freq = freq;
+	attr.freq = 1;
```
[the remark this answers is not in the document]
ThinkerYzu1: It switches between sample_period & sample_freq.  According to the man page and my understanding, sample_period is a fixed interval., for example, the number of ticks.  sample_freq specified a frequency, kernel will adjust the sampling period to achieve the desired rate.  For example, CPU clocks may change.

examples/c/profile.h
```diff
@@ -0,0 +1,22 @@
...
+#define MAX_STACK_DEPTH         127
+#endif
+
+typedef __u64 stack_trace_t[MAX_STACK_DEPTH];
+
+struct stacktrace_event {
+	__u32 pid;
+	__u32 cpu_id;
+	__u32 kernel_stack_sz;
+	__u32 user_stack_sz;
```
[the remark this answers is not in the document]
ThinkerYzu1: I will set them to 0 if bpf_get_stack() fails.

examples/rust/bzsstack/src/main.rs
ThinkerYzu1: I just found a problem with RingBuffer API.  RingBuffer::poll() accepts only Duration.  While ring_buffer__poll() waits indefinitely if no event happens, for passing -1 as the value of timeout, Duration can not be negative.

examples/c/profile.c
anakryiko: it's useful to preserve error code, actually, so I'd leave the error code and report it in user-space (if it happens). BTW, I'm not sure what happens when perf_event is triggered during user-space code execution (and thus there shouldn't be a kernel stack)?

examples/c/profile.c
anakryiko: yep, just use __u64 consistently everywhere

examples/c/profile.c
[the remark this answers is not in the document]
anakryiko: hm, but with HW_CPU_CYCLES we do want fixed amount of CPU cycles to pass, it shouldn't be adjusted by OS, it's a hardware feature. So should we just use a .sample_period instead? I think that's also what FB's internal profiler is using.

examples/c/profile.c
```diff
@@ -0,0 +1,197 @@
...
+
+cleanup:
+	if (links) {
+		for (cpu = 0; cpu < num_cpus; cpu++)
+			bpf_link__destroy(links[cpu]);
+		free(links);
+	}
+	if (pefds) {
+		for (i = 0; i < num_cpus; i++)
+			if (pefds[i] >= 0)
```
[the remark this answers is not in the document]
anakryiko: it's practically impossible to have valid perf FD == 0 in our case, so I wouldn't bother, but nothing wrong with -1 initialization, if you insist

examples/c/profile.h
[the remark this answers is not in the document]
anakryiko: see above, I think preserving error code is better, it's not hard to handle that in user-space and provide extra information if getting stack trace failed

examples/rust/bzsstack/src/main.rs
anakryiko: bootstrap example specifies 100ms as a timeout, it doesn't hurt. But @insearchoflosttime, probably something to consider supporting to specify "no timeout" ?

insearchoflosttime: Yes, we should support that for RingBuffer::poll(). I will make sure that gets fixed in libbpf-rs

examples/rust/bzsstack/src/main.rs
```diff
@@ -0,0 +1,211 @@
...
+    };
+    attr._type = syscall::PERF_TYPE_HARDWARE;
+    attr.size = mem::size_of::<syscall::perf_event_attr>() as u32;
+    attr.config = syscall::PERF_COUNT_HW_CPU_CYCLES;
+    attr.sample.sample_freq = freq;
+    attr.flags = 1 << 10; // freq = 1
+    (0..nprocs)
+        .map(|cpu| {
+            let fd = syscall::perf_event_open(attr.as_ref(), pid, cpu as i32, -1, 0);
+            fd as libc::c_int
```
insearchoflosttime, line 58: Is there a reason this is returned as libc::c_int instead of i32? It seems like returning this as a vector of i32s will make things easier (eg in attach_perf_event, perfds can be an &[i32] and not have to do any casting)?

insearchoflosttime:
This looks great btw! i tried out bzsstack locally and it looks awesome. One question I have is what printing out the "stack[i] - sym.start_address" is helpful for. For example for something like:

```
PID 0, CPU 0
Kernel:
  0 [<ffffffff81100368>] update_blocked_averages+0x38 /ro/source/kernel/sched/sched.h:1312
  1 [<ffffffff8110099f>] update_nohz_stats+0x3f /ro/source/kernel/sched/fair.c:8575
...
```
what does the +0x38 and +0x3f signify?

anakryiko:
> what does the +0x38 and +0x3f signify?

it allows to know exact assembly instruction in that function, which is useful in a lot of different contexts. One example:
```
$ gdb ~/linux-build/default/vmlinux
...
(gdb) list *(update_blocked_averages+0x38)
0xffffffff810a56d8 is in update_blocked_averages (/data/users/andriin/linux/kernel/sched/sched.h:1521).
1516    static inline void rq_pin_lock(struct rq *rq, struct rq_flags *rf)
1517    {
1518            rf->cookie = lockdep_pin_lock(__rq_lockp(rq));
1519
1520    #ifdef CONFIG_SCHED_DEBUG
1521            rq->clock_update_flags &= (RQCF_REQ_SKIP|RQCF_ACT_SKIP);
1522            rf->clock_update_flags = 0;
1523    #ifdef CONFIG_SMP
1524            SCHED_WARN_ON(rq->balance_callback && rq->balance_callback != &balance_push_callback);
1525    #endif
```

Pretty nice, right?

examples/rust/bzsstack/src/bpf/profile.h
```diff
@@ -0,0 +1,22 @@
+/* SPDX-License-Identifier: (LGPL-2.1 OR BSD-2-Clause) */
+/* Copyright (c) 2022 Meta Platforms, Inc. */
+#ifndef __PROFILE_H_
+#define __PROFILE_H_
+
+#ifndef MAX_STACK_DEPTH
+#define MAX_STACK_DEPTH         127
```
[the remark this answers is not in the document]
ThinkerYzu1: It was 128.  But due to PERF_SAMPLE_CALLCHAIN of perf_event_open() returns N+1 addresses, it is 127 now even though I don't use PERF_SAMPLE_CALLCHAIN anymore.  I should change it to 128 now.

examples/rust/bzsstack/src/main.rs
ThinkerYzu1: You are right!  The reason for casting to c_int is not valid anymore.

ThinkerYzu1:
I just updated the patches with some major changes.

-  Make freq adjustable. 
- Capture and show COMM of processes as well.
- Use symlink instead of copies of BPF source codes.

I don't remove the definitions of perf_event_attr from the rust code since the latest libbpf-rs still depends on libbpf-sys having no full definition of perf_event_attr.

examples/rust/bzsstack/src/main.rs
ThinkerYzu1: Use Duration::MAX instead.

anakryiko:
Great work and great examples, thanks! I had to add profile to .gitignore (please pay attention to stuff like that for future submissions), and also had to add `-lm` to LD_FLAGS in examples/c/Makefile due to:
```
/opt/rh/gcc-toolset-11/root/usr/bin/ld: /data/users/andriin/libbpf-bootstrap/examples/c/.output/libblazesym.a(std-7c582493123fc1dd.std.f40690d7-cgu.0.rcgu.o): in function `std::f32::<impl f32>::lerp':
/rustc/f1edd0429582dd29cccacaf50fd134b05593bd9c//library/std/src/f32.rs:914: undefined reference to `fmaf'
/opt/rh/gcc-toolset-11/root/usr/bin/ld: /data/users/andriin/libbpf-bootstrap/examples/c/.output/libblazesym.a(std-7c582493123fc1dd.std.f40690d7-cgu.0.rcgu.o): in function `std::f64::<impl f64>::lerp':
/rustc/f1edd0429582dd29cccacaf50fd134b05593bd9c//library/std/src/f64.rs:916: undefined reference to `fma'
```

In terms of the output I have few nits, if you can follow up on them it would be great. E.g.,:

```
  9 [<ffffffff81bda69d>] do_syscall_64+0x45
  10 [<ffffffff81c00068>] entry_SYSCALL_64_after_hwframe+0x44 /ro/source/arch/x86/entry/entry_64.S:112
```

I think it would be good to make sure that both 9 and 10 are right-aligned and padded with spaces, so output doesn't shift at 10th frame. And can you add () around source line information (it's a bit hard to see it)? So output is more like:

```
   9 [<ffffffff81bda69d>] do_syscall_64+0x45
  10 [<ffffffff81c00068>] entry_SYSCALL_64_after_hwframe+0x44 (/ro/source/arch/x86/entry/entry_64.S:112)
```

I saw some examples of non-symbolized frames and some errors (something about "Fail to ceate ElfResolver ..."), I'll follow up with on for these offline.

anakryiko:
Oh, one more thing. Can you please add a brief description of C/Rust profile example to README, just like we do that for all other examples?

PR #80, opened 2022-05-28
Add USDT example

chenhengqi:
Add a new example to demonstrate the usage of libbpf USDT functionality.

### DIFF
--- a/README.md
+++ b/README.md
@@ -115,6 +115,28 @@ $ sudo cat /sys/kernel/debug/tracing/trace_pipe
            <...>-461101 [018] d... 505434.345367: bpf_trace_printk: UPROBE EXIT: return = 5
 ```
 
+## USDT
+
+`usdt` is an example of dealing with USDT probe. It attaches USDT BPF programs to
+the [libc:setjmp](https://www.gnu.org/software/libc/manual/html_node/Non_002dlocal-Goto-Probes.html) probe, which is triggered by calling `setjmp` in user-space program once per second and logs USDT arguments using `bpf_printk()` macro:
+
+```shell
+$ sudo ./usdt
+libbpf: loading object 'usdt_bpf' from buffer
+...
+Successfully started!
+...........
+```
+
+You can see `usdt` demo output in `/sys/kernel/debug/tracing/trace_pipe`:
+```shell
+$ sudo cat /sys/kernel/debug/tracing/trace_pipe
+            usdt-1919077 [005] d..21 537310.886092: bpf_trace_printk: USDT auto attach to libc:setjmp: arg1 = 55d03d6a42a0, arg2 = 0, arg3 = 55d03d65e54e
+            usdt-1919077 [005] d..21 537310.886105: bpf_trace_printk: USDT manual attach to libc:setjmp: arg1 = 55d03d6a42a0, arg2 = 0, arg3 = 55d03d65e54e
+            usdt-1919077 [005] d..21 537311.886214: bpf_trace_printk: USDT auto attach to libc:setjmp: arg1 = 55d03d6a42a0, arg2 = 0, arg3 = 55d03d65e54e
+            usdt-1919077 [005] d..21 537311.886227: bpf_trace_printk: USDT manual attach to libc:setjmp: arg1 = 55d03d6a42a0, arg2 = 0, arg3 = 55d03d65e54e
+```
+
 # Fentry
 
 `fentry` is an example that uses fentry and fexit BPF programs for tracing. It
--- a/examples/c/usdt.bpf.c
+++ b/examples/c/usdt.bpf.c
@@ -1,0 +1,29 @@
+// SPDX-License-Identifier: (LGPL-2.1 OR BSD-2-Clause)
+/* Copyright (c) 2022 Hengqi Chen */
+#include <vmlinux.h>
+#include <bpf/bpf_helpers.h>
+#include <bpf/bpf_tracing.h>
+#include <bpf/usdt.bpf.h>
+
+pid_t my_pid;
+
+SEC("usdt/libc.so.6:libc:setjmp")
+int BPF_USDT(usdt_auto_attach, void *arg1, int arg2, void *arg3)
+{
+	pid_t pid = bpf_get_current_pid_tgid() >> 32;
+
+	if (pid != my_pid)
+		return 0;
+
+	bpf_printk("USDT auto attach to libc:setjmp: arg1 = %lx, arg2 = %d, arg3 = %lx", arg1, arg2, arg3);
+	return 0;
+}
+
+SEC("usdt/libc.so.6:libc:setjmp")
+int BPF_USDT(usdt_manual_attach, void *arg1, int arg2, void *arg3)
+{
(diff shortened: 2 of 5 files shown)


README.md
```diff
@@ -96,6 +96,29 @@ $ sudo cat /sys/kernel/debug/tracing/trace_pipe
            <...>-461101 [018] d... 505434.345367: bpf_trace_printk: UPROBE EXIT: return = 5
 ```
 
+## USDT
+
+`usdt` is an example of dealing with USDT probe which requires `systemtap-sdt-devel`
```
anakryiko, line 101: I'm a bit worried that we require extra package here.  How about instead of defining our own USDT, example program would be attaching to existing USDTs in some well-known binary or shared library? Can you check if glibc, or bash, or something that's with high likelihood is going to be present on end user's system has some USDTs we can attach to?

anakryiko: @chenhengqi  any updates on this?

chenhengqi: Updated. PTAL.

examples/c/usdt.bpf.c
```diff
@@ -0,0 +1,15 @@
+// SPDX-License-Identifier: (LGPL-2.1 OR BSD-2-Clause)
+/* Copyright (c) 2022 Hengqi Chen */
+#include <vmlinux.h>
+#include <bpf/bpf_helpers.h>
+#include <bpf/bpf_tracing.h>
+#include <bpf/usdt.bpf.h>
+
+SEC("usdt//lib/x86_64-linux-gnu/libc.so.6:libc:setjmp")
```
anakryiko, line 8: can we use `libc.so.6` and not specify absolute path? libbpf has extra logic to try to find the library in common search path, so by relying on that this example can hopefully work on non-x86 architectures as well

examples/c/usdt.bpf.c
```diff
@@ -0,0 +1,15 @@
...
+int BPF_USDT(usdt_test, void *arg1, int arg2, void *arg3)
+{
+	bpf_printk("USDT libc:setjmp: arg1 = %x, arg2 = %d, arg3 = %x", arg1, arg2, arg3);
```
anakryiko, line 11: should it be %lx for arg1 and arg3?

examples/c/usdt.c
```diff
@@ -0,0 +1,102 @@
...
+		if (sscanf(filename, "libc-%f.so", &version) == 1) {
+			memcpy(path, buf, strlen(buf));
+			fclose(f);
+			return 0;
+		}
+	}
+
+	fclose(f);
+	return -1;
+}
```
anakryiko, line 48: you shouldn't need this

examples/c/usdt.c
```diff
@@ -0,0 +1,102 @@
...
+		return 1;
+	}
+
+	err = get_libc_path(libc_path);
+	if (err) {
+		fprintf(stderr, "Failed to get libc path\n");
+		goto cleanup;
+	}
+
+	skel->links.usdt_test = bpf_program__attach_usdt(skel->progs.usdt_test, getpid(),
```
anakryiko, line 75: the way you specified BPF_USDT() program's SEC() it should be auto-attachable, so you can use iether usdt_bpf__attach() for skeleton or bpf_program__attach() for one specific program. But it would be good to demonstrate both auto-attach capabilities and have an example of explicit attach like you do here. So maybe let's have two BPF programs (I think it's fine if it's for the same USDT probe, just let's have two different bpf_printk() messages).

But either way you shouldn't need to specify absolute path for libc (though do test this, of course). Thanks!

anakryiko (changes requested):

chenhengqi:
@anakryiko Sorry for the long delay. I've just updated this example. Please take a look.

anakryiko:
@chenhengqi I did some further simplification and pushed straight to master. Thanks for working on this! BTW, if you get a chance before I get there, it would be nice to also adjust uprobe example to show case libbpf's support for library resolution (like in this case below, specifying just "libc.so.6" is enough) and attachment to functions by name. It's a pretty recent feature that a lot of people probably don't know about.

Anyways, changes I did:

```diff
diff --git a/examples/c/usdt.bpf.c b/examples/c/usdt.bpf.c
index 4cd792f8d3d3..418aa481fca4 100644
--- a/examples/c/usdt.bpf.c
+++ b/examples/c/usdt.bpf.c
@@ -19,7 +19,7 @@ int BPF_USDT(usdt_auto_attach, void *arg1, int arg2, void *arg3)
 	return 0;
 }
 
-SEC("usdt/libc.so.6:libc:setjmp")
+SEC("usdt")
 int BPF_USDT(usdt_manual_attach, void *arg1, int arg2, void *arg3)
 {
 	bpf_printk("USDT manual attach to libc:setjmp: arg1 = %lx, arg2 = %d, arg3 = %lx", arg1, arg2, arg3);
diff --git a/examples/c/usdt.c b/examples/c/usdt.c
index d501c94f1ec1..345238c7d302 100644
--- a/examples/c/usdt.c
+++ b/examples/c/usdt.c
@@ -7,8 +7,6 @@
 #include "usdt.skel.h"
 
 static volatile sig_atomic_t exiting;
-static const char *usdt_provider = "libc";
-static const char *usdt_name = "setjmp";
 static jmp_buf env;
 
 static void sig_int(int signo)
@@ -21,39 +19,12 @@ static int libbpf_print_fn(enum libbpf_print_level level, const char *format, va
 	return vfprintf(stderr, format, args);
 }
 
-static int get_libc_path(char *path)
-{
-	FILE *f;
-	char buf[PATH_MAX] = {};
-	char *filename;
-	float version;
-
-	f = fopen("/proc/self/maps", "r");
-	if (!f)
-		return -errno;
-
-	while (fscanf(f, "%*x-%*x %*s %*s %*s %*s %[^\n]\n", buf) != EOF) {
-		if (strchr(buf, '/') != buf)
-			continue;
-		filename = strrchr(buf, '/') + 1;
-		if (sscanf(filename, "libc-%f.so", &version) == 1) {
-			memcpy(path, buf, strlen(buf));
-			fclose(f);
-			return 0;
-		}
-	}
-
-	fclose(f);
-	return -1;
-}
-
 static void usdt_trigger() {
 	setjmp(env);
 }
 
 int main(int argc, char **argv)
 {
-	char libc_path[PATH_MAX] = {};
 	struct usdt_bpf *skel;
 	int err;
 
@@ -74,18 +45,12 @@ int main(int argc, char **argv)
 		return 1;
 	}
 
-	err = get_libc_path(libc_path);
-	if (err) {
-		fprintf(stderr, "Failed to get libc path\n");
-		goto cleanup;
-	}
-
 	/*
 	 * Manually attach to libc.so we find.
 	 * We specify pid here, so we don't have to do pid filtering in BPF program.
 	 */
 	skel->links.usdt_manual_attach = bpf_program__attach_usdt(skel->progs.usdt_manual_attach, getpid(),
-								  libc_path, usdt_provider, usdt_name, NULL);
+								  "libc.so.6", "libc", "setjmp", NULL);
 	if (!skel->links.usdt_manual_attach) {
 		err = errno;
 		fprintf(stderr, "Failed to attach BPF program `usdt_manual_attach`\n");
```

chenhengqi:
That's cool. Will do.

Issue #68, opened 2022-04-24
minimal example can't run


baoqger:
After compile and build the `minimal` example application, I try to run it but with the following error message:

```
libbpf: loading object 'minimal_bpf' from buffer
libbpf: elf: section(2) tp/syscalls/sys_enter_write, size 104, link 0, flags 6, type=1
libbpf: sec 'tp/syscalls/sys_enter_write': found program 'handle_tp' at insn offset 0 (0 bytes), code size 13 insns (104 bytes)
libbpf: elf: section(3) license, size 13, link 0, flags 3, type=1
libbpf: license of minimal_bpf is Dual BSD/GPL
libbpf: elf: section(4) .bss, size 4, link 0, flags 3, type=8
libbpf: elf: section(5) .rodata, size 28, link 0, flags 2, type=1
libbpf: elf: section(6) .BTF, size 617, link 0, flags 0, type=1
libbpf: elf: section(7) .BTF.ext, size 160, link 0, flags 0, type=1
libbpf: elf: section(8) .symtab, size 216, link 13, flags 0, type=2
libbpf: elf: section(9) .reltp/syscalls/sys_enter_write, size 32, link 8, flags 0, type=9
libbpf: looking for externs among 9 symbols...
libbpf: collected 0 externs total
libbpf: map 'minimal_.bss' (global data): at sec_idx 4, offset 0, flags 400.
libbpf: map 0 is "minimal_.bss"
libbpf: map 'minimal_.rodata' (global data): at sec_idx 5, offset 0, flags 480.
libbpf: map 1 is "minimal_.rodata"
libbpf: sec '.reltp/syscalls/sys_enter_write': collecting relocation for section(2) 'tp/syscalls/sys_enter_write'
libbpf: sec '.reltp/syscalls/sys_enter_write': relo #0: insn #2 against 'my_pid'
libbpf: prog 'handle_tp': found data map 0 (minimal_.bss, sec 4, off 0) for insn 2
libbpf: sec '.reltp/syscalls/sys_enter_write': relo #1: insn #6 against '.rodata'
libbpf: prog 'handle_tp': found data map 1 (minimal_.rodata, sec 5, off 0) for insn 6
libbpf: Kernel doesn't support BTF, skipping uploading it.
libbpf: prog 'handle_tp': relo #0: kernel doesn't support global data
libbpf: prog 'handle_tp': failed to relocate data references: -95
libbpf: failed to load object 'minimal_bpf'
[...]

PR #84, opened 2022-06-08
Add new minimal_legacy example

Darrilla:
As requested in #68 discussion, this PR adds a new minimal_legacy example for running on kernels that do not support global variables.

### DIFF
--- a/examples/c/minimal_legacy.bpf.c
+++ b/examples/c/minimal_legacy.bpf.c
@@ -1,0 +1,34 @@
+/* SPDX-License-Identifier: (LGPL-2.1 OR BSD-2-Clause) */
+#define BPF_NO_GLOBAL_DATA
+#include <linux/bpf.h>
+#include <bpf/bpf_helpers.h>
+#include <bpf/bpf_tracing.h>
+
+typedef unsigned int u32;
+typedef int pid_t;
+
+char LICENSE[] SEC("license") = "Dual BSD/GPL";
+
+/* Create an array with 1 entry instead of a global variable
+ * which does not work with older kernels */
+struct {
+	__uint(type, BPF_MAP_TYPE_ARRAY);
+	__uint(max_entries, 1);
+	__type(key, u32);
+	__type(value, pid_t);
+} my_pid_map SEC(".maps");
+
+SEC("tp/syscalls/sys_enter_write")
+int handle_tp(void *ctx)
+{
+	u32 index = 0;
+	pid_t pid = bpf_get_current_pid_tgid() >> 32;
+	pid_t *my_pid = bpf_map_lookup_elem(&my_pid_map, &index);
+
+	if (!my_pid || *my_pid != pid)
+		return 1;
+
+	bpf_printk("BPF triggered from PID %d.\n", pid);
+
+	return 0;
+}
--- a/examples/c/minimal_legacy.c
+++ b/examples/c/minimal_legacy.c
@@ -1,0 +1,58 @@
+/* SPDX-License-Identifier: (LGPL-2.1 OR BSD-2-Clause) */
+#include <stdio.h>
+#include <unistd.h>
+#include <sys/resource.h>
+#include <bpf/libbpf.h>
+#include "minimal_legacy.skel.h"
+
+static int libbpf_print_fn(enum libbpf_print_level level, const char *format, va_list args)
+{
+	return vfprintf(stderr, format, args);
+}
+
+int main(int argc, char **argv)
+{
+	struct minimal_legacy_bpf *skel;
+	int err;
+	pid_t pid;
+	unsigned index = 0;
+
+	libbpf_set_strict_mode(LIBBPF_STRICT_ALL);
+	/* Set up libbpf errors and debug info callback */
+	libbpf_set_print(libbpf_print_fn);
+
+	/* Load and verify BPF application */
+	skel = minimal_legacy_bpf__open_and_load();
+	if (!skel) {
+		fprintf(stderr, "Failed to open and load BPF skeleton\n");
+		return 1;
+	}
+
+	/* ensure BPF program only handles write() syscalls from our process */
+	pid = getpid();
+	err = bpf_map__update_elem(skel->maps.my_pid_map, &index, sizeof(index), &pid, sizeof(pid_t), BPF_ANY);
+	if (err < 0) {
+		fprintf(stderr, "Error updating map with pid: %s\n", strerror(err));
+		goto cleanup;
+	}
+
+	/* Attach tracepoint handler */
+	err = minimal_legacy_bpf__attach(skel);
+	if (err) {
+		fprintf(stderr, "Failed to attach BPF skeleton\n");
+		goto cleanup;
+	}
+
+	printf("Successfully started! Please run `sudo cat /sys/kernel/debug/tracing/trace_pipe` "
+	       "to see output of the BPF programs.\n");
+
+	for (;;) {
+		/* trigger our BPF program */
+		fprintf(stderr, ".");
+		sleep(1);
+	}
+
+cleanup:
+	minimal_legacy_bpf__destroy(skel);
+	return -err;
+}
--- a/README.md
+++ b/README.md
@@ -20,6 +20,25 @@ $ sudo cat /sys/kernel/debug/tracing/trace_pipe
 `minimal` is great as a bare-bones experimental playground to quickly try out
 new ideas or BPF features.
 
+## Minimal_Legacy
+
+This version of `minimal` is modified to allow running on even older kernels
+that do not allow global variables. bpf_printk uses global variables unless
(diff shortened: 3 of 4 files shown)


examples/c/minimal_legacy.bpf.c
```diff
@@ -0,0 +1,32 @@
+// SPDX-License-Identifier: GPL-2.0 OR BSD-3-Clause
+/* Copyright (c) 2020 Facebook */
```
anakryiko, line 2: outdated and wrong Copyright?

examples/c/minimal_legacy.bpf.c
```diff
@@ -0,0 +1,32 @@
...
+#define BPF_NO_GLOBAL_DATA
+#include "vmlinux.h"
+#include <bpf/bpf_helpers.h>
+#include <bpf/bpf_tracing.h>
+#include <bpf/bpf_core_read.h>
```
anakryiko, line 7: we are not going to use bpf_core_Read.h, drop the include?

examples/c/minimal_legacy.bpf.c
```diff
@@ -0,0 +1,32 @@
+// SPDX-License-Identifier: GPL-2.0 OR BSD-3-Clause
+/* Copyright (c) 2020 Facebook */
+#define BPF_NO_GLOBAL_DATA
+#include "vmlinux.h"
```
anakryiko, line 4: given we are not using CO-RE (and legacy systems won't have vmlinux BTF), let's include UAPI headers we need instead? Line linux/bpf.h

examples/c/minimal_legacy.bpf.c
```diff
@@ -0,0 +1,32 @@
...
+#include <bpf/bpf_helpers.h>
+#include <bpf/bpf_tracing.h>
+#include <bpf/bpf_core_read.h>
+
+char LICENSE[] SEC("license") = "Dual BSD/GPL";
+
+// Create an array with 1 entry instead of a global variable
```
anakryiko, line 11: please don't use C++-style comments, stick to C89

examples/c/minimal_legacy.bpf.c
```diff
@@ -0,0 +1,32 @@
...
+	__type(value, pid_t);
+} my_pid SEC(".maps");
+
+SEC("tp/syscalls/sys_enter_write")
+int handle_tp(void *ctx)
+{
+	u32 index = 0;
+	pid_t pid = bpf_get_current_pid_tgid() >> 32;
+	pid_t *user_pid = bpf_map_lookup_elem(&my_pid, &index);
+	if (user_pid)
```
anakryiko, line 26: empty line after variable declaration block (C89 style)

examples/c/minimal_legacy.bpf.c
```diff
@@ -0,0 +1,32 @@
...
+	{
+		if (*user_pid == pid)
```
anakryiko, line 28: combine two ifs and invert the condition to exit early?

```
if (user_pid && *user_pid == pid)
    return 1;

bpf_printk(...);
return 0;
```

examples/c/minimal_legacy.c
```diff
@@ -0,0 +1,62 @@
...
+
+	/* Load and verify BPF application */
+	skel = minimal_legacy_bpf__open();
+	if (!skel) {
+		fprintf(stderr, "Failed to open BPF skeleton\n");
+		return 1;
+	}
+
+	/* Load & verify BPF programs */
+	err = minimal_legacy_bpf__load(skel);
```
anakryiko, line 33: you can combine open and load into single `minimal_legacy_bpf__open_and_load()`, keep it shorter

examples/c/minimal_legacy.c
```diff
@@ -0,0 +1,62 @@
...
+	if (err) {
+		fprintf(stderr, "Failed to load and verify BPF skeleton\n");
+		goto cleanup;
+	}
+
+	/* ensure BPF program only handles write() syscalls from our process */
+	pid = getpid();
+	bpf_map__update_elem(skel->maps.my_pid, &index, sizeof(index), &pid, sizeof(pid_t), BPF_ANY);
```
anakryiko, line 41: bonus points for using `bpf_map__update_elem()`! :)
but check the error to show a good example

examples/c/minimal_legacy.c
```diff
@@ -0,0 +1,62 @@
+// SPDX-License-Identifier: (LGPL-2.1 OR BSD-2-Clause)
+/* Copyright (c) 2020 Facebook */
```
anakryiko, line 2: copyright update?

examples/c/minimal_legacy.bpf.c
```diff
@@ -0,0 +1,32 @@
...
+char LICENSE[] SEC("license") = "Dual BSD/GPL";
+
+// Create an array with 1 entry instead of a global variable
+// which does not work with older kernels
+struct {
+	__uint(type, BPF_MAP_TYPE_ARRAY);
+	__uint(max_entries, 1);
+	__type(key, u32);
+	__type(value, pid_t);
+} my_pid SEC(".maps");
```
anakryiko, line 18: maybe call this `my_pid_map` so you can have `my_pid` local variable below? `user_pid` is a bit confusing

examples/c/minimal_legacy.bpf.c
Darrilla: What do I change this to?

examples/c/minimal_legacy.bpf.c
Darrilla: Shouldn't I still return 0? and exit early when the pid does not match:

if (user_pid && *user_pid != pid) return 0;

It may be better just to say:

if (user_pid && *user_pid == pid)
   bpf_printk(...);

examples/c/minimal_legacy.bpf.c
anakryiko: You can either drop this line altogether or use current year and your name? Not a lawyer, but that's what I've seen done in practice :)

examples/c/minimal_legacy.bpf.c
anakryiko: so 0 vs 1 in tracepoints is not about returning error, it's whether perf subsystem will generate event for tracepoint or not. Returning 0 prevents it, which can interfere with some perf uses. So returning 1 always is a better idea in general. It's just a little known detail about BPF tracepoint programs, and it usually doesn't cause problems (but sometimes it does, unfortunately).

As for exiting early, it's just cleaner because it leaves the rest of logic more linear. Anyone using this example to extend it and play with it will have easier time extending it if you exit early.

anakryiko:
BTW, please add your Signed-off-by to the commit, similar to how libbpf and kernel commits do that:

Signed-off-by: Full Name <your@email.address>

Darrilla:
I have updated this PR with requested changes.

examples/c/minimal_legacy.bpf.c
```diff
@@ -0,0 +1,34 @@
+/* SPDX-License-Identifier: (LGPL-2.1 OR BSD-2-Clause) */
+#define BPF_NO_GLOBAL_DATA
+#include "linux/bpf.h"
+#include "sched.h"
```
anakryiko, line 4: why not `#include <linux/bpf.h>` ? Also why do we need sched.h? Can we use some system-wide UAPI header instead?

examples/c/minimal_legacy.c
```diff
@@ -0,0 +1,58 @@
...
+	skel = minimal_legacy_bpf__open_and_load();
+	if (!skel) {
+		fprintf(stderr, "Failed to open and load BPF skeleton\n");
+		return 1;
+	}
+
+	/* ensure BPF program only handles write() syscalls from our process */
+	pid = getpid();
+	err = bpf_map__update_elem(skel->maps.my_pid_map, &index, sizeof(index), &pid, sizeof(pid_t), BPF_ANY);
+	if (err == -1) {
```
anakryiko, line 34: it won't be -1 in libbpf 1.0, just check `err < 0` (and you can use err for error code, no need for errno)

anakryiko (commented):
few minor things

examples/c/minimal_legacy.bpf.c
Darrilla: shed.h is for pid_t,. I should have used the <> for bpf.h. 
I could just define pid_t locally if that is easier.

Darrilla:
Try this

anakryiko (approved):
Looks good, but please squash commits into one final commit? Keep the Signed-off-by and maybe add a simple commit message (besides the subject line). Thanks!

Darrilla:
> Looks good, but please squash commits into one final commit? Keep the Signed-off-by and maybe add a simple commit message (besides the subject line). Thanks!

Done

anakryiko:
Thanks!

Issue #81, opened 2022-05-29
Xmake + Android error


neomafo88:
- host: Ubuntu 20.04.4 LTS / Kernel version 5.18-051800-generic
- guest: Android 11 x64 (with arm64 emulation by default) - stock Google Pixel 4 emulator

```
$ xmake
[ 34%]: ccache compiling.release ../../libbpf/src/libbpf.c
[ 43%]: ccache compiling.release ../../libbpf/src/usdt.c
[ 43%]: ccache compiling.release ../../libbpf/src/hashmap.c
[ 46%]: ccache compiling.release ../../libbpf/src/strset.c
error: ../../libbpf/src/usdt.c:562:7: error: unknown type name 'GElf_Nhdr'
      GElf_Nhdr *nhdr, const char *data, size_t name_off, size_t desc_off,
      ^
../../libbpf/src/usdt.c:578:11: error: expected ';' after expression
 GElf_Nhdr nhdr;
          ^
          ;
../../libbpf/src/usdt.c:578:2: error: use of undeclared identifier 'GElf_Nhdr'
 GElf_Nhdr nhdr;
 ^
../../libbpf/src/usdt.c:578:12: error: use of undeclared identifier 'nhdr'; did you mean 'ehdr'?
 GElf_Nhdr nhdr;
           ^~~~
           ehdr
../../libbpf/src/usdt.c:577:12: note: 'ehdr' declared here
 GElf_Ehdr ehdr;
           ^
  > in ../../libbpf/src/usdt.c
```

chenhengqi:
You need `libelf`.

https://github.com/libbpf/libbpf-bootstrap#install-dependencies

anakryiko:
@chenhengqi maybe it's good to mention that it has to be -dev version of libelf package? libelf itself is necessary at runtime, but for compilation you need headers which normally come with -dev/-devel (naming convention changes by Linux distro)

neomafo88:
still seems broken for me, do I have to modify the (x)makefile or something to specify some special compilation flag, or path to find includes/libs as an environment variable?

![screenshot](https://user-images.githubusercontent.com/97744946/172911372-f6db181f-3451-4e47-b873-b642e86e9ac4.png)

waruqi:
The new version of libbpf uses GElf_Nhdr, but the libelf library in xmake-repo is version 0.8.13, which does not provide the GElf_Nhdr type.

[...]

PR #85, opened 2022-06-18
fix compile errors for android

waruqi:


ThinkerYzu1:
@anakryiko Do we want to switch to elfutils from libelf?

waruqi:
This patch is only to fix https://github.com/libbpf/libbpf-bootstrap/issues/81 because xmake uses the old libelf 0.8.13, which does not provide the GElf_Nhdr type, but the new libbpf library uses the GElf_Nhdr Nhdr.

Currently only the libelf library inside elfutils provides GElf_Nhdr

ThinkerYzu1:
Since we actually need libelf instead of elfutils, it would be better to depend on libelf.

FYI, here is the way to fix reversion issues with xmake.
https://xmake.io/#/package/remote_package?id=install-any-version-of-the-package

waruqi:
> Since we actually need libelf instead of elfutils, it would be better to depend on libelf.

The libelf-dev package for debian is actually from the elfutils library.

The original libelf library is, https://directory.fsf.org/wiki/Libelf, which is no longer maintained and is stuck at 0.8.13

Therefore, the libelf library on which the libbpf library relies is always elfutils.

ThinkerYzu1:
@waruqi Thank you for the clarification.  LGTM!

PR #90, opened 2022-07-02
Update uprobe example

chenhengqi:


examples/c/uprobe.c
```diff
@@ -78,15 +83,15 @@ int main(int argc, char **argv)
 	 * parse ELF to calculate this function, we'd need to add .text
 	 * section offset and function's offset within .text ELF section.
 	 */
-	uprobe_offset = (long)&uprobed_function - base_addr;
+	uprobe_offset = (long)&uprobed_add - base_addr;
```
anakryiko, line 86: This is (generally speaking) broken, btw (I've only found it out relatively recently), see https://github.com/libbpf/libbpf/commit/2d91c46d1af8678c1ee626100b167405b5ea7447 for some details on how newer linkers can generate more complicated ELF and VMA arrangements.

We've actually addressed this in selftests/bpf (see https://github.com/torvalds/linux/blob/master/tools/testing/selftests/bpf/trace_helpers.c#L149), it would be great to change uprobe example approach to that one (but maybe skip powerpc64 quirks for simplicity?)

examples/c/uprobe.c
```diff
@@ -96,24 +101,31 @@ int main(int argc, char **argv)
...
+							       -1 /* any pid */,
+							       "/proc/self/exe",
+							       uprobe_offset);
+	if (!skel->links.uretprobe_add) {
 		err = -errno;
 		fprintf(stderr, "Failed to attach uprobe: %d\n", err);
 		goto cleanup;
 	}
 
+	err = uprobe_bpf__attach(skel);
```
anakryiko, line 115: how about instead showing off generic per-program attachment here:

```
skel_.links.uprobe_sub = bpf_program__attach(skel->progs.uprobe_add);
```
and the same for retprobe (with error checking, of course).

I think a lot of newbies will glance over uprobe_bpf__attach() and won't realize what's going on and where and when uprobe attachment is actually happening.

WDYT?

examples/c/uprobe.c
ThinkerYzu1: `dlinfo(handle, RTLD_DI_LINKMAP, ..)` should be easier for this case.  `l_addr` in `struct link_map` is the base address you need.

anakryiko: oh, is l_addr a file offset or virtaddr offset, do you know?

but either way, the point of uprobe example is to show how you realistically will get file offset for some function in *any* binary, not necessarily own process, so dlinfo() is too limiting in this case. I think having /proc/<pid>/maps approach demonstrated is a better example in this case

anakryiko:
@chenhengqi friendly ping, will you get a chance to do those last small changes? Thanks!

examples/c/uprobe.c
chenhengqi: The `uprobe_bpf__attach` is to demostrate auto-attach, just like the USDT example.

anakryiko:
Sorry for delay, thanks for updates!

PR #97, opened 2022-07-31
examples/c: add socket filter example

JackyYin:
This is a simplified example I port from linux `samples/bpf` directory, many code snippet came from there.

In this example, I opened a `SOCK_RAW` socket in `lo` interface, then attach it to the bpf program for monitoring.

I use `BPF_MAP_TYPE_RINGBUF` to carry information back to user space program, and the user space program will print formatted information to stdout, for example:
```
interface:lo    protocol: UDP   127.0.0.1:51845(src) -> 127.0.0.1:53(dst)
interface:lo    protocol: UDP   127.0.0.1:41552(src) -> 127.0.0.1:53(dst)
```

Currently, most of the IPv4 protocols defined in `uapi/linux/in.h` are included, please check `ipproto_mapping` for the supported protocols.

Another thing to be done is that we need to add another linux uapi header into `libbpf` submodule :  `uapi/linux/ip.h`, because we need to retrieve `protocol`, `saddr` and `daddr` from `struct ip_hdr`.

I will temporarily mark this PR as draft, please give me some suggestions for this, thank you!

### DIFF
--- a/README.md
+++ b/README.md
@@ -254,8 +254,8 @@ please check `ipproto_mapping` of `examples/c/sockfilter.c` for the supported pr
 
 ```shell
 $ sudo ./sockfilter
-protocol: UDP   127.0.0.1:51845(src) -> 127.0.0.1:53(dst)
-protocol: UDP   127.0.0.1:41552(src) -> 127.0.0.1:53(dst)
+interface:lo    protocol: UDP   127.0.0.1:51845(src) -> 127.0.0.1:53(dst)
+interface:lo    protocol: UDP   127.0.0.1:41552(src) -> 127.0.0.1:53(dst)
 ```
 
 # Building
--- a/README.md
+++ b/README.md
@@ -252,16 +252,11 @@ protocol, src IP, src port, dst IP, dst port in standard output.
 Currently, most of the IPv4 protocols defined in `uapi/linux/in.h` are included,
 please check `ipproto_mapping` of `examples/c/sockfilter.c` for the supported protocols.
 
-
-
 ```shell
 $ sudo ./sockfilter
-protocol: UDP
-127.0.0.1:51845(src) -> 127.0.0.1:53(dst)
-protocol: UDP
-127.0.0.1:41552(src) -> 127.0.0.1:53(dst)
+protocol: UDP   127.0.0.1:51845(src) -> 127.0.0.1:53(dst)
+protocol: UDP   127.0.0.1:41552(src) -> 127.0.0.1:53(dst)
 ```
-
 
 # Building
 
--- a/README.md
+++ b/README.md
@@ -137,7 +137,7 @@ $ sudo cat /sys/kernel/debug/tracing/trace_pipe
             usdt-1919077 [005] d..21 537311.886227: bpf_trace_printk: USDT manual attach to libc:setjmp: arg1 = 55d03d6a42a0, arg2 = 0, arg3 = 55d03d65e54e
 ```
 
-# Fentry
+## Fentry
 
 `fentry` is an example that uses fentry and fexit BPF programs for tracing. It
 attaches `fentry` and `fexit` traces to `do_unlinkat()` which is called when a
@@ -172,7 +172,7 @@ $ sudo cat /sys/kernel/debug/tracing/trace_pipe
               rm-9290    [004] d..2  4637.798843: bpf_trace_printk: fexit: pid = 9290, filename = test_file2, ret = 0
 ```
 
-# Kprobe
+## Kprobe
 
 `kprobe` is an example of dealing with kernel-space entry and exit (return)
 probes, `kprobe` and `kretprobe` in libbpf lingo. It attaches `kprobe` and
@@ -198,7 +198,7 @@ $ sudo cat /sys/kernel/debug/tracing/trace_pipe
               rm-9346    [005] d..4  4710.951895: bpf_trace_printk: KPROBE EXIT: ret = 0
 ```
 
-# XDP
+## XDP
 
 `xdp` is an example written in Rust (using libbpf-rs). It attaches to
 the ingress path of networking device and logs the size of each packet,
@@ -221,7 +221,7 @@ $ sudo cat /sys/kernel/debug/tracing/trace_pipe
            <...>-2813507 [000] d.s1 602386.696735: bpf_trace_printk: packet size: 66
 ```
 
-# Profile
+## Profile
 
 `profile` is an example written in Rust and C with BlazeSym. It
 attaches to perf events, sampling on every processor periodically. It
@@ -242,6 +242,26 @@ No Userspace Stack
 ```
 
 C version and Rust version show the same content.  Both of them use BlazeSym to symbolize stacktraces.
+
+## Socket filter
+
+`sockfilter` is an example of monitoring packet and dealing with `__sk_buff`
+structure. It attaches `socket` BPF program to `sock_queue_rcv_skb()` function
+and retrieve information from `BPF_MAP_TYPE_RINGBUF`, then print
+protocol, src IP, src port, dst IP, dst port in standard output.
+Currently, most of the IPv4 protocols defined in `uapi/linux/in.h` are included,
+please check `ipproto_mapping` of `examples/c/sockfilter.c` for the supported protocols.
+
+
+
+```shell
+$ sudo ./sockfilter
+protocol: UDP
+127.0.0.1:51845(src) -> 127.0.0.1:53(dst)
+protocol: UDP
+127.0.0.1:41552(src) -> 127.0.0.1:53(dst)
+```
+
 
 # Building
 
--- a/examples/c/sockfilter.bpf.c
+++ b/examples/c/sockfilter.bpf.c
@@ -1,0 +1,73 @@
+// SPDX-License-Identifier: GPL-2.0 OR BSD-3-Clause
+/* Copyright (c) 2022 Jacky Yin */
+#include <linux/bpf.h>
+#include <linux/if_ether.h>
+#include <linux/ip.h>
+#include <linux/in.h>
+#include <bpf/bpf_helpers.h>
+#include <bpf/bpf_endian.h>
+#include <stddef.h>
+#define IP_MF		0x2000
+#define IP_OFFSET	0x1FFF
+
+char LICENSE[] SEC("license") = "Dual BSD/GPL";
+
+struct {
+	__uint(type, BPF_MAP_TYPE_RINGBUF);
+	__uint(max_entries, 256 * 1024);
+} rb SEC(".maps");
+
+struct so_event {
+	__be32 src_addr;
+	__be32 dst_addr;
+	union {
+		__be32 ports;
+		__be16 port16[2];
+	};
+	__u32 ip_proto;
+	__u32 pkt_type;
+};
+
+static inline int ip_is_fragment(struct __sk_buff *skb, __u32 nhoff)
+{
+	__u16 frag_off;
+	bpf_skb_load_bytes(skb, nhoff + offsetof(struct iphdr, frag_off), &frag_off, 2);
+	frag_off = __bpf_ntohs(frag_off);
+	return frag_off & (IP_MF | IP_OFFSET);
+}
+
+SEC("socket")
+int socket_handler(struct __sk_buff *skb)
+{
+	struct so_event *e;
+	__u8 verlen;
+	__u16 proto;
+	__u32 nhoff = ETH_HLEN;
+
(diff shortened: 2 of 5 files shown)


README.md
```diff
@@ -243,6 +243,26 @@ No Userspace Stack
...
+
+`sockfilter` is an example of monitoring packet and dealing with `__sk_buff`
+structure. It attaches `socket` BPF program to `sock_queue_rcv_skb()` function
+and retrieve information from `BPF_MAP_TYPE_RINGBUF`, then print
+protocol, src IP, src port, dst IP, dst port in standard output.
+Currently, most of the IPv4 protocols defined in `uapi/linux/in.h` are included,
+please check `ipproto_mapping` of `examples/c/sockfilter.c` for the supported protocols.
+
+
+
```
anakryiko, line 256: nit: why so many empty lines? :)

README.md
```diff
@@ -243,6 +243,26 @@ No Userspace Stack
...
+```shell
+$ sudo ./sockfilter
+protocol: UDP
+127.0.0.1:51845(src) -> 127.0.0.1:53(dst)
+protocol: UDP
+127.0.0.1:41552(src) -> 127.0.0.1:53(dst)
+```
+
+
```
anakryiko, line 265: nit: ditto

examples/c/sockfilter.bpf.c
```diff
@@ -0,0 +1,73 @@
+// SPDX-License-Identifier: GPL-2.0 OR BSD-3-Clause
+/* Copyright (c) 2022 Jacky Yin */
+#include <linux/bpf.h>
+#include <linux/if_ether.h>
+#include <linux/ip.h>
+#include <linux/in.h>
+#include <bpf/bpf_helpers.h>
+#include <bpf/bpf_endian.h>
+#include <stddef.h>
+#define IP_MF		0x2000
```
anakryiko, line 10: nit: please add empty line above to separate includes from defines

examples/c/sockfilter.bpf.c
```diff
@@ -0,0 +1,73 @@
...
+	__be32 dst_addr;
+	union {
+		__be32 ports;
+		__be16 port16[2];
+	};
+	__u32 ip_proto;
+	__u32 pkt_type;
+};
+
+static inline int ip_is_fragment(struct __sk_buff *skb, __u32 nhoff)
```
anakryiko, line 31: does it need to be inline?

examples/c/sockfilter.bpf.c
```diff
@@ -0,0 +1,73 @@
...
+{
+	__u16 frag_off;
```
anakryiko, line 33: empty line after variable declaration section

examples/c/sockfilter.c
```diff
@@ -0,0 +1,177 @@
...
+#include <linux/if_ether.h>
+#include <linux/in.h>
+#include <net/if.h>
+#include <stdio.h>
+#include <sys/resource.h>
+#include <sys/socket.h>
+#include <unistd.h>
+#include "sockfilter.skel.h"
+
+struct so_event {
```
anakryiko, line 16: this struct is used in BPF part and in user-space part, let's extract it into a common header, just like bootstrap.h does?

examples/c/sockfilter.c
```diff
@@ -0,0 +1,177 @@
...
+};
+
+static inline int open_raw_sock(const char *name) {
+	struct sockaddr_ll sll;
+	int sock;
+
+	sock = socket(PF_PACKET, SOCK_RAW | SOCK_NONBLOCK | SOCK_CLOEXEC,
+				htons(ETH_P_ALL));
+	if (sock < 0) {
+		printf("cannot create raw socket\n");
```
anakryiko, line 62: given it's an error, let's use fprintf(stderr) instead?

examples/c/sockfilter.c
```diff
@@ -0,0 +1,177 @@
...
+		return -1;
+	}
+
+	memset(&sll, 0, sizeof(sll));
+	sll.sll_family = AF_PACKET;
+	sll.sll_ifindex = if_nametoindex(name);
+	sll.sll_protocol = htons(ETH_P_ALL);
+	if (bind(sock, (struct sockaddr *)&sll, sizeof(sll)) < 0) {
+		printf("bind to %s: %s\n", name, strerror(errno));
```
anakryiko, line 71: nit: "failed to bind to ..." ? and see above about fprintf(stderr)

examples/c/sockfilter.c
```diff
@@ -0,0 +1,177 @@
...
+static int handle_event(void *ctx, void *data, size_t data_sz) {
+	const struct so_event *e = data;
+
+	if (e->pkt_type != PACKET_HOST)
+		return 0;
+
+	if (e->ip_proto < 0 || e->ip_proto >= IPPROTO_MAX)
+		return 0;
+
+	printf("protocol: %s\n%s:%d(src) -> %s:%d(dst)\n",
```
anakryiko, line 94: nit: maybe keep one record per single line:

proto: %s\t%s:%d (src) -> %s:%d (dst)\n

?

examples/c/sockfilter.c
```diff
@@ -0,0 +1,177 @@
...
+	return sock;
+}
+
+static int libbpf_print_fn(enum libbpf_print_level level, const char *format, va_list args)
+{
+	return vfprintf(stderr, format, args);
+}
+
+
+static int handle_event(void *ctx, void *data, size_t data_sz) {
```
anakryiko, line 85: let's keep styling consistent with kernel style, { should be on next line (same in other places)

examples/c/sockfilter.c
```diff
@@ -0,0 +1,177 @@
...
+
+	/* Open BPF application */
+	skel = sockfilter_bpf__open();
+	if (!skel) {
+		fprintf(stderr, "Failed to open BPF skeleton\n");
+		return 1;
+	}
+
+	/* Load & verify BPF programs */
+	err = sockfilter_bpf__load(skel);
```
anakryiko, line 121: you can combine open and load steps into one: sockfilter_bpf__open_and_load();

examples/c/sockfilter.c
```diff
@@ -0,0 +1,177 @@
...
+	prog_fd = bpf_program__fd(skel->progs.socket_handler);
+	sock = open_raw_sock("lo");
+	if (sock < 0) {
+		err = -2;
+		fprintf(stderr, "Failed to open raw socket\n");
+		goto cleanup;
+	}
+
+	if (setsockopt(sock, SOL_SOCKET, SO_ATTACH_BPF, &prog_fd,
+		sizeof(prog_fd))) {
```
anakryiko, line 152: not sure if it's Github UI or formatting is off, but please make sure that sizeof() is aligned with sock (or given it's not too long the line, just keep all this on a single line?

iamkafai:
In general, have you thought about doing the same thing in tc hook instead?  tc is a more popular hook for bpf prog.  Also, it should be faster (kernel does not always have to clone the skb) and has more bpf features/helpers.  For example, it can directly read skb->data.  The tc hook has much more helper to find useful things also.  Just name a few, when the tc-bpf prog intercepted an interesting src/dst ip, it can find the sk and print more information about this sk.  e.g. the sock cookie and what tcp state it is in...etc.

README.md
JackyYin: Fixed it!

README.md
JackyYin: Fixed it too!

examples/c/sockfilter.bpf.c
JackyYin: Fixed it too.

examples/c/sockfilter.bpf.c
JackyYin: Sure, fixed it.

examples/c/sockfilter.c
JackyYin: OK, now chnaged to something like this:
`fprintf(stderr, "Failed to bind to %s: %s\n", name, strerror(errno));`

examples/c/sockfilter.c
JackyYin: Sure, fixed it!

examples/c/sockfilter.c
JackyYin: No problem, fixed all the related functions.

examples/c/sockfilter.c
JackyYin: OK, now I use fprintf(stderr) for all of the error messages.

examples/c/sockfilter.c
JackyYin: It's a good idea.
I added a new socketfilter.h file.

examples/c/sockfilter.bpf.c
JackyYin: Actually I don't think it need to be inline, but I though it would be a little bit better for performance.
Do you have any concern about this?

examples/c/sockfilter.c
JackyYin: Thanks for reminding me about this. Fixed!

examples/c/sockfilter.c
JackyYin: It's not because of the UI.
I merged it into the single line now.

examples/c/sockfilter.bpf.c
```diff
@@ -0,0 +1,73 @@
...
+		bpf_skb_load_bytes(skb, nhoff + offsetof(struct iphdr, saddr), &(e->src_addr), 4);
+		bpf_skb_load_bytes(skb, nhoff + offsetof(struct iphdr, daddr), &(e->dst_addr), 4);
+	}
+
+	bpf_skb_load_bytes(skb, nhoff + 0, &verlen, 1);
+	bpf_skb_load_bytes(skb, nhoff + ((verlen & 0xF) << 2), &(e->ports), 4);
+	e->pkt_type = skb->pkt_type;
+	bpf_ringbuf_submit(e, 0);
+
+	return skb->len;
```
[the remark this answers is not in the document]
JackyYin: Actually, I found out that I can still receive packet from userspace program even if I return 0 in the very beginning of the eBPF program, so I think return value doesn't matter.

JackyYin:
> In general, have you thought about doing the same thing in tc hook instead? tc is a more popular hook for bpf prog. Also, it should be faster (kernel does not always have to clone the skb) and has more bpf features/helpers. For example, it can directly read skb->data. The tc hook has much more helper to find useful things also. Just name a few, when the tc-bpf prog intercepted an interesting src/dst ip, it can find the sk and print more information about this sk. e.g. the sock cookie and what tcp state it is in...etc.

Thanks a bunch for all of these useful information!
Actually I haven't tried this before, but I think we can cover as many bpf program types as possible:
https://github.com/libbpf/libbpf/blob/4eb6485c08867edaa5a0a81c64ddb23580420340/src/libbpf.c#L9002-L9081
Maybe I can add another example for `tc` bpf program later, what kind of implementation do you think, is suitable for a `tc` bpf program?

anakryiko:
Sorry for delay, looks good. skeleton__attach() isn't necessary (socket programs are not auto-attachable, so it has no effect, I'll drop it and force-push locally, no need to resubmit).

But also (I'm not a big BPF networking expert), at least for some use cases, skb->data and skb->data_end are probably a better approach than more heavyweight bpf_skb_load_bytes(). Have you used data/data_end special pointers before? Do they work in this case? If yes, maybe let's use them at least for some accesses to improve the example and demonstrate both ways of accessing data?

JackyYin:
> Sorry for delay, looks good. skeleton__attach() isn't necessary (socket programs are not auto-attachable, so it has no effect, I'll drop it and force-push locally, no need to resubmit).
> 
> But also (I'm not a big BPF networking expert), at least for some use cases, skb->data and skb->data_end are probably a better approach than more heavyweight bpf_skb_load_bytes(). Have you used data/data_end special pointers before? Do they work in this case? If yes, maybe let's use them at least for some accesses to improve the example and demonstrate both ways of accessing data?

After tracking down linux source code, I thought skb->data and skb->data_end are not accessible in BPF_PROG_TYPE_SOCKET_FILTER, reference:
https://github.com/torvalds/linux/blob/072e51356cd5a4a1c12c1020bc054c99b98333df/net/core/filter.c#L8365-L8392

I will try to use this in other examples, thanks a lot for your suggestions!

anakryiko:
Very well could be, as I said not a big BPF networking expert here :)

JackyYin:
> Very well could be, as I said not a big BPF networking expert here :)

It's OK, you looks like an expert to me!
Still thanks for the information anyway.

PR #103, opened 2022-08-25
examples/c: Add TC example

chenhengqi:
Add a TC example to showcase the libbpf TC APIs.
Also serve as an answer to question raised in this discussion ([0]).

  [0]: https://github.com/libbpf/libbpf-bootstrap/discussions/101

Signed-off-by: Hengqi Chen <chenhengqi@outlook.com>

### DIFF
--- a/examples/c/tc.bpf.c
+++ b/examples/c/tc.bpf.c
@@ -1,0 +1,34 @@
+// SPDX-License-Identifier: (LGPL-2.1 OR BSD-2-Clause)
+/* Copyright (c) 2022 Hengqi Chen */
+#include <vmlinux.h>
+#include <bpf/bpf_endian.h>
+#include <bpf/bpf_helpers.h>
+#include <bpf/bpf_tracing.h>
+
+#define TC_ACT_OK	0
+#define ETH_P_IP	0x0800		/* Internet Protocol packet	*/
+
+SEC("tc")
+int tc_ingress(struct __sk_buff *ctx)
+{
+	void *data_end = (void *)(__u64)ctx->data_end;
+	void *data = (void *)(__u64)ctx->data;
+	struct ethhdr *l2;
+	struct iphdr *l3;
+
+	if (ctx->protocol != bpf_htons(ETH_P_IP))
+		return TC_ACT_OK;
+
+	l2 = data;
+	if ((void *)(l2 + 1) > data_end)
+		return TC_ACT_OK;
+
+	l3 = (struct iphdr *)(l2 + 1);
+	if ((void *)(l3 + 1) > data_end)
+		return TC_ACT_OK;
+
+	bpf_printk("Got IP packet: tot_len: %d, ttl: %d", bpf_ntohs(l3->tot_len), l3->ttl);
+	return TC_ACT_OK;
+}
+
+char __license[] SEC("license") = "GPL";
--- a/examples/c/tc.c
+++ b/examples/c/tc.c
@@ -1,0 +1,87 @@
+// SPDX-License-Identifier: (LGPL-2.1 OR BSD-2-Clause)
+/* Copyright (c) 2022 Hengqi Chen */
+#include <signal.h>
+#include <unistd.h>
+#include "tc.skel.h"
+
+#define LO_IFINDEX	1
+
+static volatile sig_atomic_t exiting = 0;
+
+static void sig_int(int signo)
+{
+	exiting = 1;
+}
+
+static int libbpf_print_fn(enum libbpf_print_level level, const char *format, va_list args)
+{
+	return vfprintf(stderr, format, args);
+}
+
+int main(int argc, char **argv)
+{
+	DECLARE_LIBBPF_OPTS(bpf_tc_hook, tc_hook,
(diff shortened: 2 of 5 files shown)


anakryiko:
@borkmann, can you review this TC example?

@chenhengqi can you also please update README?

examples/c/tc.bpf.c
```diff
@@ -0,0 +1,34 @@
...
+SEC("tc")
+int tc_ingress(struct __sk_buff *ctx)
+{
+	void *data_end = (void *)(__u64)ctx->data_end;
+	void *data = (void *)(__u64)ctx->data;
+	struct ethhdr *l2;
+	struct iphdr *l3;
+
+	if (ctx->protocol != bpf_htons(ETH_P_IP))
+		return TC_ACT_OK;
```
vincentmli, line 20: just curious why return TC_ACT_OK to pass it to upper stack when it is not a valid IP packet? why not TC_ACT_SHOT ? I know this is just an example to show how to use TC, not how to deal with the packet so I guess it may not be  important? :)

chenhengqi: Just to avoid interrupting the system running this example. There exists other protocols like ARP.

chenhengqi:
> @borkmann, can you review this TC example?
> 
> @chenhengqi can you also please update README?

Updated.

examples/c/tc.bpf.c
vincentmli: ok make sense here, what about the following two `TC_ACT_OK`?

chenhengqi: It's used to make the verifier happy.

vincentmli: you mean this `if ((void *)(l2 + 1) > data_end)` is to make verifier happy, right? but I mean if `if` is true, the data then is  invalid, should we return `TC_ACT_SHOT`? sorry just trying to understand it better from you :)

borkmann:
> > @borkmann, can you review this TC example?
> > @chenhengqi can you also please update README?
> 
> Updated.

Looks good to me overall, the code could probably be further stripped down, e.g. to remove it to the absolute min (e.g. stripping signal handling, etc).

JackyYin:
Greeting, thanks for the great work.
Just curious, is it possible to also demonstrate how to leverage Layer 2 information in this example? (by `struct ethhdr`)?

examples/c/tc.bpf.c
cheneytianx: > you mean this `if ((void *)(l2 + 1) > data_end)` is to make verifier happy, right? but I mean if `if` is true, the data then is invalid, should we return `TC_ACT_SHOT`? sorry just trying to understand it better from you :)

If we return `TC_ACT_SHOT`, it means the current packet will be dropped. Meanwhile, the application layer receives no packet and will cause retransmission. 

Thus, the proper way to deal with the invalid packet is to return `TC_ACT_OK`, which means pass the invalid packet to the upper network stack and let the upper layer to make decision on it.

examples/c/tc.c
```diff
@@ -0,0 +1,87 @@
...
+
+static int libbpf_print_fn(enum libbpf_print_level level, const char *format, va_list args)
+{
+	return vfprintf(stderr, format, args);
+}
+
+int main(int argc, char **argv)
+{
+	DECLARE_LIBBPF_OPTS(bpf_tc_hook, tc_hook,
+		.ifindex = LO_IFINDEX, .attach_point = BPF_TC_INGRESS);
```
cheneytianx, line 24: I recommend using the following function to decide the targeted interface, since the index may vary on different devices.

```c
u_int32_t lo_ifindex = if_nametoindex("lo");
```
The function declaration is in `net/if.h`.

cheneytianx (approved):
LGTM.

examples/c/tc.c
chenhengqi: The Linux kernel ensures that loopback device is always the first network device, even in network namespace.

chenhengqi:
> Greeting, thanks for the great work. Just curious, is it possible to also demonstrate how to leverage Layer 2 information in this example? (by `struct ethhdr`)?

Users can easily retrieve these information and are expected to play with it themselves.

chenhengqi:
> > > @borkmann, can you review this TC example?
> > > @chenhengqi can you also please update README?
> > 
> > 
> > Updated.
> 
> Looks good to me overall, the code could probably be further stripped down, e.g. to remove it to the absolute min (e.g. removing signal handling, etc).

Thanks. Daniel.

examples/c/tc.bpf.c
vincentmli: got it, this lead me to the https://en.wikipedia.org/wiki/Robustness_principle "be conservative in what you do, be liberal in what you accept from others"

examples/c/tc.c
cheneytianx: > The Linux kernel ensures that loopback device is always the first network device, even in network namespace.

Got it. Thanks.

anakryiko:
@chenhengqi thanks a lot for this contribution! And thanks @borkmann for review.

PR #110, opened 2022-09-09
Fix xmake.lua 

waruqi:


anakryiko:
@waruqi sorry, can you elaborate in commit message why this is a proper fix. I find it hard to understand why removing files fixes something. It would be nice to also have your Signed-off-by in commit message, for consistency with libbpf style.

waruqi:
`add_files("minimal*.c")` will compile minimal.c and minimal_legacy.c at same time. it will cause duplicate _main conflict.

so we need remove unused `minimal_legacy.c` and `minimal_legacy.bpf.c` files in minimal target.

waruqi:
https://github.com/libbpf/libbpf-bootstrap/pull/115

PR #112, opened 2022-09-14
[ci] introduce CI to build examples against LLVM 14 to 16

chantra:
Simple CI workflow that will build both C and Rust examples.


chantra:
Example run: https://github.com/chantra/libbpf-bootstrap/actions/runs/3053576261/jobs/4924374016

While playing with libbpf-bootstrap I found out rust's tracecon failed to build with LLVM 16. I hooked up the dockerfile in agithub action so examples can be built against different llvm version in the future.

anakryiko:
This is awesome, thank you!

PR #134, opened 2022-12-05
cmake: Fix btf header missing in legacy kernel env. 

woodpenker:
If include `<bpf/btf.h>` and compile with CMake on some machines with legacy kernels, you may hit this error:
```
src/build/libbpf/bpf/btf.h:10:10: fatal error: linux/btf.h: No such file or directory
#include <linux/btf.h>
^~~~~~~~~~~~~
compilation terminated.
```
I think it's due to `../../libbpf/include/uapi` not includes as it is at [Makefile#L18](https://github.com/libbpf/libbpf-bootstrap/blob/c58877bf5895dd5015550d4624f68a6ff0ed4b33/examples/c/Makefile#L18)

After the fix, a dir containing uapi headers will be created under the `build` dir
```
$ ls build/libbpf/linux
bpf_common.h  bpf.h  btf.h                                                                        
```

### DIFF
--- a/examples/c/CMakeLists.txt
+++ b/examples/c/CMakeLists.txt
@@ -19,7 +19,7 @@ ExternalProject_Add(libbpf
     INCLUDEDIR=
     LIBDIR=
     UAPIDIR=
-    install
+    install install_uapi_headers
   BUILD_IN_SOURCE TRUE
   INSTALL_COMMAND ""
   STEP_TARGETS build


anakryiko:
Looks good, thanks. But can you please update commit message to include the description from this PR? Right now the subject itself is too brief (and we try to follow kernel style guide which says that commit messages have to have non-empty commit message). Thanks!

woodpenker:
> Looks good, thanks. But can you please update commit message to include the description from this PR? Right now the subject itself is too brief (and we try to follow kernel style guide which says that commit messages have to have non-empty commit message). Thanks!

Sure, but I also have another question: does kernel side build need this included too? It seems Makefile includes it, if so, maybe I need to add it too.

anakryiko:
This works as well. One tiny nit. Signed-off-by: is supposed to be used with real name, currently I see

Signed-off-by: woodpenker <woodpenker5@gmail.com>

Can you update woodpenker to a properly spelled full name (e.g., Andrii Nakryiko  for me). Thanks! Just for consistency with kernel and libbpf style.

anakryiko (approved):
Goodness, this fell through the cracks. I'm so sorry for the delay!

Issue #140, opened 2023-01-03
excute kprobe failed on ARM-64


LinuxVWQH:
Hi，when i installed the kprobe tools in my OS,I get a problem:
![image](https://user-images.githubusercontent.com/74600097/210333640-06dd5af0-beba-47f0-b07b-6965cc6279f5.png)
![image](https://user-images.githubusercontent.com/74600097/210333697-c11736d8-ebb8-4f24-bed2-27077d59a7d1.png)
I try to take some measures but not useful,please help me,thanks!
**PS1**:
OS_info:
`Linux e04-g4 5.10.110-yocto-standard #1 SMP PREEMPT Tue Jan 3 09:00:28 UTC 2023 aarch64 GNU/Linux`
**PS2**:
KERNEL_CONFIG
`CONFIG_TASKS_TRACE_RCU=y
CONFIG_BPF_SYSCALL=y
CONFIG_TRACEPOINTS=y
CONFIG_UPROBES=y
CONFIG_BINARY_PRINTF=y

CONFIG_DEBUG_INFO_BTF=y
CONFIG_STACKTRACE=y
CONFIG_NOP_TRACER=y
CONFIG_RING_BUFFER=y
CONFIG_EVENT_TRACING=y
CONFIG_CONTEXT_SWITCH_TRACER=y
CONFIG_TRACING=y
CONFIG_FTRACE=y
CONFIG_FTRACE_SYSCALLS=y
CONFIG_BRANCH_PROFILE_NONE=y
CONFIG_UPROBE_EVENTS=y
CONFIG_BPF_EVENTS=y
CONFIG_DYNAMIC_EVENTS=y
CONFIG_PROBE_EVENTS=y
CONFIG_CGROUP_BPF=y
CONFIG_SOCK_CGROUP_DATA=y
CONFIG_BPF_LSM=y
CONFIG_BPF_JIT_ALWAYS_ON=y
CONFIG_USERMODE_DRIVER=y
CONFIG_KPROBES=y
CONFIG_KRETPROBES=y
CONFIG_BPF_STREAM_PARSER=y
CONFIG_STREAM_PARSER=y
CONFIG_NET_SOCK_MSG=y
CONFIG_KPROBE_EVENTS=y
CONFIG_FUNCTION_ERROR_INJECTION=y
CONFIG_DEBUG_FS=y
CONFIG_KPROBES_ON_FTRACE=y
CONFIG_HAVE_KPROBES_ON_FTRACE=y
[...]

PR #142, opened 2023-01-06
Make ARCH configurable.

ThinkerYzu1:
Doing cross-compilation (issue #140) need a way to specify the target platform.  However, the makefile also detects architecture by calling `uname` on the building machine.  This patch lets users to config it by defining ARCH.  For example, 'make ARCH=arm64' will build against arm64 as the target platform.

Signed-off-by: Kui-Feng Lee <kuifeng@meta.com>

### DIFF
--- a/examples/c/Makefile
+++ b/examples/c/Makefile
@@ -10,7 +10,7 @@ BPFTOOL ?= $(BPFTOOL_OUTPUT)/bootstrap/bpftool
 LIBBLAZESYM_SRC := $(abspath ../../blazesym/)
 LIBBLAZESYM_OBJ := $(abspath $(OUTPUT)/libblazesym.a)
 LIBBLAZESYM_HEADER := $(abspath $(OUTPUT)/blazesym.h)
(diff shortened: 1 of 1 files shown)


LinuxVWQH:
> Doing cross-compilation (issue #140) need a way to specify the target platform. However, the makefile also detects architecture by calling `uname` on the building machine. This patch lets users to config it by defining ARCH. For example, 'make ARCH=arm64' will build against arm64 as the target platform.
> 
> Signed-off-by: Kui-Feng Lee [kuifeng@meta.com](mailto:kuifeng@meta.com)

I think you should check the CMake and XMake in the meantime,because the libbpf-bootstrap support three ways:)

ThinkerYzu1:
> I think you should check the CMake and XMake in the meantime,because the libbpf-bootstrap support three ways:)

Ideally, yes!  However, I am not familiar with cmake & make.  It would be great if someone who is familiar with cmake & xmake can fix it.

anakryiko (approved):
Thanks!

PR #154, opened 2023-02-28
Use local blazesym snapshot for Rust examples

danielocfb:
We already pull in a snapshot of blazesym for the generation of C headers and the creation of a library.
Let's also use that in the Rust examples, instead of reaching out to `crates.io`, to use the same baseline everywhere.


anakryiko:
nice, thanks!

PR #161, opened 2023-03-23
examples/c: Fix Makefile for aarch64

chenhengqi:
On aarch64, C examples accidentally include arm/vmlinux.h and cause runtime error. Reorder the sed commands to avoid the issue.


anakryiko:
Nice find, thanks for the fix!

PR #163, opened 2023-03-23
examples/c: make CLANG_BPF_SYS_INCLUDES substitutable

kolerov:
Sometimes it's necessary to replace `CLANG_BPF_SYS_INCLUDES` by another value when build platform differs from host. For example, we can do something like this in Buildroot package for `libbpf-bootstrap`:

    make ... CLANG_BPF_SYS_INCLUDES="$(STAGING_DIR)/usr/include"

Otherwise, clang uses host's Linux headers instead of target's.


anakryiko (commented):
LGTM, thanks!

PR #164, opened 2023-03-24
Replace llvm-strip symlink with BPF static linking

### LINKED ISSUE #156: Build fails with llvm-strip: not found
When i run `make minimal` on Ubuntu (20.04 & 22.04) it fails with  `/bin/sh: 1: llvm-strip: not found`. In order to get this working i followed the installation https://apt.llvm.org/, and then created a symbolic link `sudo ln -s /usr/bin/llvm-strip-12 /usr/bin/llvm-strip`. Can you please improve the README.md so that others don't get hung up on this.  

`vagrant@devenv:~/host_user_homedir/workspace/src/libbpf-bootstrap/examples/c$ make minimal
  MKDIR    .output
  MKDIR    .output/libbpf
  LIB      libbpf.a
.....
  INSTALL  /home/vagrant/host_user_homedir/workspace/src/libbpf-bootstrap/examples/c/.output//libbpf/libbpf.a
  BPF      .output/minimal.bpf.o
/bin/sh: 1: llvm-strip: not found
make: *** [Makefile:112: .output/minimal.bpf.o] Error 127
make: *** Deleting file '.output/minimal.bpf.o'`

heyrutvik:
Closes #156

### DIFF
--- a/examples/c/Makefile
+++ b/examples/c/Makefile
@@ -1,7 +1,6 @@
 # SPDX-License-Identifier: (LGPL-2.1 OR BSD-2-Clause)
 OUTPUT := .output
 CLANG ?= clang
-LLVM_STRIP ?= llvm-strip
 LIBBPF_SRC := $(abspath ../../libbpf/src)
 BPFTOOL_SRC := $(abspath ../../bpftool/src)
 LIBBPF_OBJ := $(abspath $(OUTPUT)/libbpf.a)
@@ -107,10 +106,11 @@ $(LIBBLAZESYM_HEADER): $(LIBBLAZESYM_SRC)/target/release/libblazesym.a | $(OUTPU
 	$(Q)cp $(LIBBLAZESYM_SRC)/target/release/blazesym.h $@
 
 # Build BPF code
-$(OUTPUT)/%.bpf.o: %.bpf.c $(LIBBPF_OBJ) $(wildcard %.h) $(VMLINUX) | $(OUTPUT)
+$(OUTPUT)/%.bpf.o: %.bpf.c $(LIBBPF_OBJ) $(wildcard %.h) $(VMLINUX) | $(OUTPUT) $(BPFTOOL)
 	$(call msg,BPF,$@)
-	$(Q)$(CLANG) -g -O2 -target bpf -D__TARGET_ARCH_$(ARCH) $(INCLUDES) $(CLANG_BPF_SYS_INCLUDES) -c $(filter %.c,$^) -o $@
-	$(Q)$(LLVM_STRIP) -g $@ # strip useless DWARF info
+	$(eval TMP_BPF_O := $(patsubst %.bpf.o,%.tmp.bpf.o,$@))
+	$(Q)$(CLANG) -g -O2 -target bpf -D__TARGET_ARCH_$(ARCH) $(INCLUDES) $(CLANG_BPF_SYS_INCLUDES) -c $(filter %.c,$^) -o $(TMP_BPF_O)
+	$(Q)$(BPFTOOL) gen object $@ $(TMP_BPF_O)
 
 # Generate BPF skeletons
 $(OUTPUT)/%.skel.h: $(OUTPUT)/%.bpf.o | $(OUTPUT) $(BPFTOOL)


examples/c/Makefile
```diff
@@ -106,10 +105,10 @@ $(LIBBLAZESYM_HEADER): $(LIBBLAZESYM_SRC)/target/release/libblazesym.a | $(OUTPU
 	$(Q)cp $(LIBBLAZESYM_SRC)/target/release/blazesym.h $@
 
 # Build BPF code
-$(OUTPUT)/%.bpf.o: %.bpf.c $(LIBBPF_OBJ) $(wildcard %.h) $(VMLINUX) | $(OUTPUT)
+$(OUTPUT)/%.bpf.o: %.bpf.c $(LIBBPF_OBJ) $(wildcard %.h) $(VMLINUX) $(BPFTOOL) | $(OUTPUT)
```
heyrutvik, line 108: Added `$(BPFTOOL)` since `bpftool` was not found at `gen object` command.

examples/c/Makefile
```diff
@@ -106,10 +105,10 @@ $(LIBBLAZESYM_HEADER): $(LIBBLAZESYM_SRC)/target/release/libblazesym.a | $(OUTPU
...
 	$(call msg,BPF,$@)
-	$(Q)$(CLANG) -g -O2 -target bpf -D__TARGET_ARCH_$(ARCH) $(INCLUDES) $(CLANG_BPF_SYS_INCLUDES) -c $(filter %.c,$^) -o $@
-	$(Q)$(LLVM_STRIP) -g $@ # strip useless DWARF info
+	$(Q)$(CLANG) -g -O2 -target bpf -D__TARGET_ARCH_$(ARCH) $(INCLUDES) $(CLANG_BPF_SYS_INCLUDES) -c $(filter %.c,$^) -o $@.tmp
+	$(Q)$(BPFTOOL) gen object $@ $@.tmp # strip useless DWARF info
```
heyrutvik, line 111: @anakryiko I've very basic knowledge of Makefile.

Your following suggest was not working 
```
$(Q)$(BPFTOOL) gen object $@ $<
```
So I tried following to do something like `bpftool gen object exmaple.bpf.o exmaple.bpf.o` but my assumption is that because of the same name it was corrupting the object file and I was getting `libbpf: failed to get ELF header for .output/minimal.bpf.o: invalid "Elf' handle" error.  
```
$(Q)$(BPFTOOL) gen object $@ $@
``` 
So my workaround is to create temporary object file to generate final object file, e.g. `bpftool gen object exmaple.bpf.o exmaple.bpf.o.tmp` (note: first param is output file and subsequent param(s) are input files), hence `$@.tmp`.

Let me know if there is better way to handle this situation. Thanks.

anakryiko: This approach makes sense. I'd like to preserve `.o` suffix, though, so maybe let's use `%.tmp.bpf.o` as intermediate name?

anakryiko (changes requested):
Makes sense. Added few nits. Please also write commit message explaining why we are doing this. Thanks!

examples/c/Makefile
```diff
@@ -107,10 +106,10 @@ $(LIBBLAZESYM_HEADER): $(LIBBLAZESYM_SRC)/target/release/libblazesym.a | $(OUTPU
 	$(Q)cp $(LIBBLAZESYM_SRC)/target/release/blazesym.h $@
 
 # Build BPF code
-$(OUTPUT)/%.bpf.o: %.bpf.c $(LIBBPF_OBJ) $(wildcard %.h) $(VMLINUX) | $(OUTPUT)
+$(OUTPUT)/%.bpf.o: %.bpf.c $(LIBBPF_OBJ) $(wildcard %.h) $(VMLINUX) | $(OUTPUT) $(BPFTOOL)
 	$(call msg,BPF,$@)
-	$(Q)$(CLANG) -g -O2 -target bpf -D__TARGET_ARCH_$(ARCH) $(INCLUDES) $(CLANG_BPF_SYS_INCLUDES) -c $(filter %.c,$^) -o $@
-	$(Q)$(LLVM_STRIP) -g $@ # strip useless DWARF info
+	$(Q)$(CLANG) -g -O2 -target bpf -D__TARGET_ARCH_$(ARCH) $(INCLUDES) $(CLANG_BPF_SYS_INCLUDES) -c $(filter %.c,$^) -o $(@D)/$*.tmp.bpf.o
+	$(Q)$(BPFTOOL) gen object $@ $(@D)/$*.tmp.bpf.o
```
anakryiko, line 112: how about replacing "$(@D)/$*.tmp.bpf.o" with "$($@:.bpf.o=.tmp.bpf.o)". Seems cleaner and more straightforward?

heyrutvik: @anakryiko, thank you for suggesting a better alternative. I'm facing an issue though.
I found out from the [doc](https://makefiletutorial.com/#string-substitution) that the expression you suggested is shorthand for the `patsubst` function call.
In our case, `$(patsubst %.bpf.o,%.tmp.bpf.o,$@)` is working fine but not `$($@:.bpf.o=.tmp.bpf.o)` (returns empty value). I'm not sure why! Any thoughts?

anakryiko: ChatGPT says it's how make works :) let's use `$(patsubst)`, no big deal!

anakryiko: my main concern is the use of `$(@D)/$*` which are less frequently used and less familiar expressions, so let's do patsubst with `$@`

anakryiko (approved):
I don't think we need `$(TMP_BPF_O)` variable, especially with `$(eval)`, I'll inline that locally while applying. Thanks!

Issue #106, opened 2022-09-04
ksyscall example


grubeli:
Libbpf v1 introduced `ksyscall` option to trace syscalls, an example for the feature would be helpful

ThinkerYzu1:
There are a lot of test cases under tools/testing/selftests/bpf directory of Linux.  You may want to check there for examples that you don't find here.

FYI
https://github.com/torvalds/linux/blob/master/tools/testing/selftests/bpf/progs/test_probe_user.c

anakryiko:
@ThinkerYzu1 yes, selftests can serve as an example, but having minimal self-contained and simple examples in libbpf-bootstrap is extremely helpful. I do agree that we should either add a separate ksyscall example or extend kprobe one to also show ksyscall one. So let's keep this issue for now until someone finds time to contribute such an example.

anakryiko:
we have ksyscall example now: https://github.com/libbpf/libbpf-bootstrap/blob/master/examples/c/ksyscall.bpf.c

PR #167, opened 2023-04-01
examples/c: Add ksyscall example

syogaraj:
Adding an example for ksyscall as said in #106

### DIFF
--- a/examples/c/ksyscall.bpf.c
+++ b/examples/c/ksyscall.bpf.c
@@ -1,0 +1,50 @@
+#include "vmlinux.h"
+#include <bpf/bpf_helpers.h>
+#include <bpf/bpf_tracing.h>
+#include <bpf/bpf_core_read.h>
+
+#define TASK_COMM_LEN 16
+
+SEC("ksyscall/tgkill")
+int BPF_KSYSCALL(tgkill_entry, pid_t tgid, pid_t tid, int sig) 
+{
+	char comm[TASK_COMM_LEN];
+
+	__u32 caller_pid = bpf_get_current_pid_tgid() >> 32;
+	if (sig == 0) {
+		/*
+			If sig is 0, then no signal is sent, but existence and permission
+			checks are still performed; this can be used to check for the
+			existence of a process ID or process group ID that the caller is
+			permitted to signal.
+		*/
+		return 0;
+	}
+
+	bpf_get_current_comm(&comm, sizeof(comm));
+	bpf_printk("tgkill syscall called by PID %d (%s) for thread id %d with pid %d and signal %d.", caller_pid, comm, tid, tgid, sig);
+	return 0;
+}
+
+SEC("ksyscall/kill")
+int BPF_KSYSCALL(entry_probe, pid_t pid, int sig) 
+{
+	char comm[TASK_COMM_LEN];
+
+	__u32 caller_pid = bpf_get_current_pid_tgid() >> 32;
+	if (sig == 0) {
+		/*
+			If sig is 0, then no signal is sent, but existence and permission
+			checks are still performed; this can be used to check for the
+			existence of a process ID or process group ID that the caller is
+			permitted to signal.
+		*/
+		return 0;
+	}
+
+	bpf_get_current_comm(&comm, sizeof(comm));
+	bpf_printk("KILL syscall called by PID %d (%s) for PID %d with signal %d.", caller_pid, comm, pid, sig);
+	return 0;
+}
+
+char _license[] SEC("license") = "GPL";
--- a/examples/c/.gitignore
+++ b/examples/c/.gitignore
@@ -9,3 +9,4 @@
 /usdt
 /sockfilter
 /tc
+/ksyscall
--- a/examples/c/Makefile
+++ b/examples/c/Makefile
@@ -24,7 +24,7 @@ INCLUDES := -I$(OUTPUT) -I../../libbpf/include/uapi -I$(dir $(VMLINUX))
 CFLAGS := -g -Wall
 ALL_LDFLAGS := $(LDFLAGS) $(EXTRA_LDFLAGS)
 
-APPS = minimal minimal_legacy bootstrap uprobe kprobe fentry usdt sockfilter tc
+APPS = minimal minimal_legacy bootstrap uprobe kprobe fentry usdt sockfilter tc ksyscall
 
 CARGO ?= $(shell which cargo)
 ifeq ($(strip $(CARGO)),)
--- a/examples/c/ksyscall.c
+++ b/examples/c/ksyscall.c
@@ -1,0 +1,60 @@
+#include <stdio.h>
+#include <unistd.h>
+#include <signal.h>
+#include <string.h>
+#include <errno.h>
+#include <sys/resource.h>
+#include <bpf/libbpf.h>
+#include "ksyscall.skel.h"
+
+static int libbpf_print_fn(enum libbpf_print_level level, const char *format, va_list args)
+{
+	return vfprintf(stderr, format, args);
+}
+
+static volatile sig_atomic_t stop;
+
+static void sig_int(int signo)
+{
+	stop = 1;
+}
+
+int main(int argc, char **argv)
+{
+	struct ksyscall_bpf *skel;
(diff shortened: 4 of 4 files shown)


chenhengqi (changes requested):
Except the inline comments, the subject line of your commit message should be `examples/c: Add ksyscall example`.

syogaraj:
@chenhengqi , have made the suggested changes. please review.

examples/c/ksyscall.bpf.c
```diff
@@ -0,0 +1,31 @@
...
+#include <bpf/bpf_tracing.h>
+#include <bpf/bpf_core_read.h>
+
+#define TASK_COMM_LEN 16
+
+SEC("ksyscall/kill")
+int BPF_KSYSCALL(entry_probe, pid_t tpid, int sig) 
+{
+
+	char comm[TASK_COMM_LEN];
```
chenhengqi, line 12: `char comm[TASK_COMM_LEN] = {}`, let's zero initialize here.

examples/c/ksyscall.bpf.c
```diff
@@ -0,0 +1,31 @@
...
+			If sig is 0, then no signal is sent, but existence and permission
+			checks are still performed; this can be used to check for the
+			existence of a process ID or process group ID that the caller is
+			permitted to signal.
+		*/
+		return 0;
+	}
+
+	if(bpf_get_current_comm(&comm, sizeof(comm)))
+		comm[0] = ' ';
```
chenhengqi, line 24: Just do `bpf_get_current_comm(&comm, sizeof(comm));`.

examples/c/ksyscall.bpf.c
```diff
@@ -0,0 +1,30 @@
...
+#include <bpf/bpf_tracing.h>
+#include <bpf/bpf_core_read.h>
+
+#define TASK_COMM_LEN 16
+
+SEC("ksyscall/kill")
+int BPF_KSYSCALL(entry_probe, pid_t tpid, int sig) 
+{
+
+	char comm[TASK_COMM_LEN] = {};
```
chenhengqi, line 12: Add a newline below, see other examples. And no need for the above empty line.

chenhengqi (commented):
Please add an entry to `.gitignore`, other than that, LGTM.

examples/c/ksyscall.bpf.c
```diff
@@ -0,0 +1,28 @@
...
+#include <bpf/bpf_helpers.h>
+#include <bpf/bpf_tracing.h>
+#include <bpf/bpf_core_read.h>
+
+#define TASK_COMM_LEN 16
+
+SEC("ksyscall/kill")
+int BPF_KSYSCALL(entry_probe, pid_t tpid, int sig) 
+{
+	char comm[TASK_COMM_LEN] = {};
```
anakryiko, line 11: there is no need to `= {}` initialize, `bpf_get_current_comm()` will ensure it's initialized

anakryiko (changes requested):
Looks good, thanks! Please squash all the commits into a single final commit, we don't need 7 commits for one example. Thanks!

examples/c/ksyscall.bpf.c
```diff
@@ -0,0 +1,27 @@
...
+		/**
+			If sig is 0, then no signal is sent, but existence and permission
+			checks are still performed; this can be used to check for the
+			existence of a process ID or process group ID that the caller is
+			permitted to signal.
+		*/
+		return 0;
+	}
+
+	bpf_get_current_comm(&comm, sizeof(comm));
```
anakryiko, line 22: you lost `comm` variable somewhere along the way? see https://github.com/libbpf/libbpf-bootstrap/actions/runs/4643677178/jobs/8233653034?pr=167

examples/c/ksyscall.bpf.c
```diff
@@ -0,0 +1,27 @@
...
+	bpf_printk("KILL syscall called on %d (%s) with signal: %d.", tpid, comm, sig);
```
anakryiko, line 23: so I get that KILL was requested for `tpid`, but it wasn't really requested by `comm`, right? It's quite misleading. comm is from current process that is issuing kill() syscall. So let's make it a bit clearer. And also log pid of the calling process, just for the fun of it?

Something like: "KILL syscall called by PID %d (%s) for PID %d"

examples/c/ksyscall.bpf.c
```diff
@@ -0,0 +1,27 @@
+#include "vmlinux.h"
+#include <bpf/bpf_helpers.h>
+#include <bpf/bpf_tracing.h>
+#include <bpf/bpf_core_read.h>
+
+#define TASK_COMM_LEN 16
+
+SEC("ksyscall/kill")
+int BPF_KSYSCALL(entry_probe, pid_t tpid, int sig)
```
anakryiko, line 9: what's tpid? manpage says it's always process ID and is called `pid`. If you'd like to trace thread kill syscalls, there is `tgkill()` syscall. Perhaps let's update the example to trace both and distinguish between them. Seems like a bit more fun example?

anakryiko (changes requested):
this doesn't compile, please test a bit more thoroughly. And see other comments and suggestions.

examples/c/ksyscall.bpf.c
syogaraj: Apologies! Instead of removing the initialization, I've mistakenly removed the declaration also. Will get back with the changes

examples/c/ksyscall.bpf.c
```diff
@@ -0,0 +1,50 @@
...
+#include <bpf/bpf_core_read.h>
+
+#define TASK_COMM_LEN 16
+
+SEC("ksyscall/tgkill")
+int BPF_KSYSCALL(tgkill_entry, pid_t tgid, pid_t tid, int sig) 
+{
+
+    char comm[TASK_COMM_LEN];
+    __u32 caller_pid = bpf_get_current_pid_tgid() >> 32;
```
anakryiko, line 13: formatting seems off, please double check that you are not mixing spaces and tabs, use tabs consistently

also, please keep empty line between variable declarations and the rest of the code

examples/c/ksyscall.bpf.c
```diff
@@ -0,0 +1,50 @@
...
+			If sig is 0, then no signal is sent, but existence and permission
+			checks are still performed; this can be used to check for the
+			existence of a process ID or process group ID that the caller is
+			permitted to signal.
+		*/
+		return 0;
+	}
+
+	bpf_get_current_comm(&comm, sizeof(comm));
+    bpf_printk("tgkill syscall called by PID %d (%s) for thread id %d with signal %d.", caller_pid, comm, tid, sig);
```
anakryiko, line 25: same here, UI renders it misaligned, please double check

anakryiko: also log both TID and PID, given tgkill has to specify both?

examples/c/ksyscall.bpf.c
```diff
@@ -0,0 +1,50 @@
...
+	return 0;
+}
+
+SEC("ksyscall/kill")
+int BPF_KSYSCALL(entry_probe, pid_t pid, int sig) 
+{
+
+    char comm[TASK_COMM_LEN];
+    __u32 caller_pid = bpf_get_current_pid_tgid() >> 32;
+	if (sig == 0) {
```
anakryiko, line 35: ditto about indentation consistency and empty line

examples/c/ksyscall.bpf.c
```diff
@@ -0,0 +1,50 @@
...
+		/**
```
anakryiko, line 36: no need to use `/**`, this is not a doc comment, just `/*`. It's also kernel coding style to use the following style:

```
/* Start of
 * multi-line
 * comment
 */
```

let's keep consistent styling

anakryiko (changes requested):
left few comments about formatting and code style. I also noticed that your commit message only has first line and no description, please add commit message, even if a simple sentence saying that we add an example demonstrating libbpf's `SEC("ksyscall")` use by tracing `tgkill()` and `kill()` syscalls.

Thanks!

PR #175, opened 2023-05-01
Minimal Example fix for pid problem in process namespaces envs like containers or WSL2

unikzforce:
Currently minimal example does not work in environments which are inside a process namespace, like docker containers or WSL2, because the pid of the `minimal` process inside the userspace is not equal to `tgid` of that process in the bpf.

this is the fix.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,6 +22,23 @@ $ sudo cat /sys/kernel/debug/tracing/trace_pipe
 `minimal` is great as a bare-bones experimental playground to quickly try out
 new ideas or BPF features.
 
+## Minimal_ns
+
+`minimal_ns` is as same as `minimal` but for namespaced environments.
+`minimal` would not work in environments that have namespace, like containers, or WSL2, 
+Cause the perceived pid of the process in the namespace is not the actual pid of the
+process. for executing `minimal` in namespaced environments like containers or wsl2 you
+need to use `minimal_namespaced` instead.
+
+```shell
+$ cd examples/c
+$ make minimal_ns
+$ sudo ./minimal_ns
+$ sudo cat /sys/kernel/debug/tracing/trace_pipe
+           <...>-3840345 [022] d...1  8804.331204: bpf_trace_printk: BPF triggered from PID 9087.
+           <...>-3840345 [022] d...1  8804.331215: bpf_trace_printk: BPF triggered from PID 9087.
+```
+
 ## Minimal_Legacy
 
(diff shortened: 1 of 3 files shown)


unikzforce:
Or maybe I should create another example side by side `minimal` like `minimal_container` so that `minimal` would remain untouched and simple and people also could learn how to prepare for namespace environment by seeing `minimal_container`.

wdyt?

anakryiko:
Yes, let's keep minimal as is, and add a new example demonstrating namespaced pid

zouyonghao:
This is so great as I just meet this issue with WSL2!

unikzforce:
Sorry guys, I was under heavy workload in my current job, I'll fix this PR very soon.

unikzforce:
@anakryiko could you please review again?

README.md
```diff
@@ -22,6 +22,11 @@ $ sudo cat /sys/kernel/debug/tracing/trace_pipe
 `minimal` is great as a bare-bones experimental playground to quickly try out
 new ideas or BPF features.
 
+`minimal` would not work in environments that have namespace, like containers, or WSL2,
```
anakryiko, line 25: see Minimal_Legacy below, let's keep each binary as a separate section?

examples/c/minimal_namespaced.bpf.c
```diff
@@ -0,0 +1,25 @@
+// SPDX-License-Identifier: GPL-2.0 OR BSD-3-Clause
+/* Copyright (c) 2020 Facebook */
```
anakryiko, line 2: update copyright year and company/author name?

examples/c/minimal_namespaced.c
```diff
@@ -0,0 +1,65 @@
+// SPDX-License-Identifier: (LGPL-2.1 OR BSD-2-Clause)
+/* Copyright (c) 2020 Facebook */
```
anakryiko, line 2: ditto

anakryiko (changes requested):
Looks good overall, few nits in the comments.

But let's fix up commit formatting:
 - use "examples/c: " prefix in commit subject
 - it's probably to name commit something more like "add minimal_ns example demonstrating PID namespace translation"
 - Please fill out commit description, not just subject. It could be very minimal, but it has to be present.
 - Signed-off-by: tag requires full name to be used. I presume "unikzforce" is not your full name, right?

And yes, "minimal_namespaced" is pretty mouthful, let's shorten it to "minimal_ns"?

PR #178, opened 2023-05-03
Avoid profiling offline or not present CPUs

tzssangglass:
In the current implementation, if I try to run `example/c/profile` in a virtual machine, `libbpf_num_possible_cpus` will get the wrong number of CPUs, causing the program to exit and throw an error: `Fail to set up performance monitor on a CPU/Core` because the wrong CPU id is passed to `syscall` and throwing `No such device`

The Rust version (`example/rust/profile`) of the implementation ignores this error and continues to execute `bpf_program__attach_perf_event`, which throws `invalid perf event FD -1`, but profile can still be used because `bpf_program__attach_perf_event handles` the case of passing the wrong CPU id.

This fix has only been applied to the C(`example/c/profile`).

### DIFF
--- a/examples/c/profile.c
+++ b/examples/c/profile.c
@@ -3,6 +3,7 @@
 #include <stdio.h>
 #include <unistd.h>
 #include <string.h>
+#include <fcntl.h>
 #include <sys/syscall.h>
 #include <sys/sysinfo.h>
 #include <linux/perf_event.h>
@@ -123,6 +124,9 @@ int main(int argc, char * const argv[])
 	int num_cpus;
 	int *pefds = NULL, pefd;
 	int argp, i, err = 0;
+	const char *online_cpus_file = "/sys/devices/system/cpu/online";
+	bool *online = NULL;
+	int num_online_cpus;
 
 	while ((argp = getopt(argc, argv, "hf:")) != -1) {
 		switch (argp) {
@@ -137,6 +141,12 @@ int main(int argc, char * const argv[])
 			show_help(argv[0]);
 			return 1;
 		}
+	}
+
+	err = parse_cpu_mask_file(online_cpus_file, &online, &num_online_cpus);
+	if (err) {
+		fprintf(stderr, "Fail to get online CPU numbers: %d\n", err);
+		goto cleanup;
 	}
 
 	num_cpus = libbpf_num_possible_cpus();
@@ -179,6 +189,10 @@ int main(int argc, char * const argv[])
 	attr.freq = 1;
 
 	for (cpu = 0; cpu < num_cpus; cpu++) {
+		/* skip offline/not present CPUs */
+		if ((cpu >= num_online_cpus || !online[cpu]))
+			continue;
+
 		/* Set up performance monitoring on a CPU/Core */
 		pefd = perf_event_open(&attr, pid, cpu, -1, PERF_FLAG_FD_CLOEXEC);
 		if (pefd < 0) {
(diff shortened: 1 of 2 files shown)


examples/c/profile.c
```diff
@@ -113,6 +114,85 @@ static void show_help(const char *progname)
...
+	}
+	return 0;
+cleanup:
+	free(*mask);
+	*mask = NULL;
+	return err;
+}
+
+// copy from https://github.com/libbpf/libbpf/blob/v1.2.0/src/libbpf.c#L12470
+int bootstrap_parse_cpu_mask_file(const char *fcpu, bool **mask, int *mask_sz)
```
anakryiko, line 169: make it static? and then you don't need to prepend `bootstrat_` prefix, just leave original name?

examples/c/profile.c
```diff
@@ -113,6 +114,85 @@ static void show_help(const char *progname)
 	printf("Usage: %s [-f <frequency>] [-h]\n", progname);
 }
 
+// copy from https://github.com/libbpf/libbpf/blob/v1.2.0/src/libbpf.c#L12420
+int bootstrap_parse_cpu_mask_str(const char *s, bool **mask, int *mask_sz)
```
anakryiko, line 118: make it static? and please use C comments (`/* */`)

examples/c/profile.c
```diff
@@ -139,6 +222,12 @@ int main(int argc, char * const argv[])
 		}
 	}
 
+	err = bootstrap_parse_cpu_mask_file(online_cpus_file, &online, &num_online_cpus);
```
anakryiko, line 225: make sure that if parsing succeeds, you free `online` mask in cleanup

anakryiko (changes requested):
few small issues, please fix

anakryiko:
Also, please write a commit message, not just subject. It doesn't have to be extensive, but should explain the purpose of the commit. Thanks!

anakryiko:
An one more thing, presumably "tzssangglass" isn't your real name. `Signed-off-by: ` requires full real name to be used, so please update that as well. thanks!

examples/c/profile.c
tzssangglass: done

tzssangglass:
> An one more thing, presumably "tzssangglass" isn't your real name. `Signed-off-by: ` requires full real name to be used, so please update that as well. thanks!

I have already modified the `Signed-off-by`, now it is my real name. I don't know why the DCO check failed.

anakryiko:
Ok, just few tiny things.

1. Please squash two commits into one final one.
2. Please make sure to wrap long lines in commit message to be under 80 characters
3. I guess I'm fine with calling into `parse_cpu_mask_file` directly to make code shorter and because we know that we link statically to libbpf. But please add comment next to extern that this function is coming from libbpf, but is not a public API and we can do this only as a demonstration and because we statically link against libbpf built from submodule. Ok?

Otherwise looks great, thanks!

tzssangglass:
> * Please squash two commits into one final one.
> * Please make sure to wrap long lines in commit message to be under 80 characters
> * I guess I'm fine with calling into `parse_cpu_mask_file` directly to make code shorter and because we know that we link statically to libbpf. But please add comment next to extern that this function is coming from libbpf, but is not a public API and we can do this only as a demonstration and because we statically link against libbpf built from submodule. Ok?

done, thank you for reviewing and providing suggestions.

anakryiko:
I did tweak your code a bit to fix formatting, cleanup, and use of user-space extern declaration in C header that's included from BPF side (we should do that). Please check the final version to see the fixes I did. Thanks!

PR #180, opened 2023-05-13
add support for clang format

unikzforce:
default ColumnLimit is 80 characters but it changed style of many files and made it ugly so I changed it to 140 for now.

for formatting the code I just have executed this command at the root of the project:
`clang-format -i ./examples/c/*.c ./examples/c/*.h`

wdyt?


examples/c/sockfilter.c
```diff
@@ -15,6 +15,7 @@
 #include "sockfilter.h"
 #include "sockfilter.skel.h"
 
+// clang-format off
```
anakryiko, line 18: what does clang-format do here that you needed to turn it off?

examples/c/uprobe.c
```diff
@@ -80,11 +80,14 @@ int main(int argc, char **argv)
 	uprobe_offset = get_uprobe_offset(&uprobed_add);
 
 	/* Attach tracepoint handler */
+	// clang-format off
```
anakryiko, line 83: well, this sucks, let's not add clang-format off in the middle of the code. What was the suggested formatting?

examples/c/uprobe.c
```diff
@@ -95,11 +98,14 @@ int main(int argc, char **argv)
 	 * processes that use the same binary executable; to do that we need
 	 * to specify -1 as PID, as we do here
 	 */
+	// clang-format off
```
anakryiko, line 101: ditto

.clang-format
```diff
@@ -0,0 +1,689 @@
...
+DisableFormat: false
+ExperimentalAutoDetectBinPacking: false
+FixNamespaceComments: false
+
+# Taken from:
+#   git grep -h '^#define [^[:space:]]*for_each[^[:space:]]*(' include/ tools/ \
+#   | sed "s,^#define \([^[:space:]]*for_each[^[:space:]]*\)(.*$,  - '\1'," \
+#   | LC_ALL=C sort -u
+ForEachMacros:
+  - '__ata_qc_for_each'
```
anakryiko, line 72: can you trim this to only keep for-each-like APIs from libbpf? There are just a few of them, no need for this long list

anakryiko (changes requested):
The changes don't look too bad, but I think 140 is too wide. The actual limit we used was 100 before, can you use that instead of 80 or 140? Let's see how that changes the code?

unikzforce:
I'll remove `// clang-format off/on` and make ColumnLimit equal to 100 , so we could compare what would be the outcome.

anakryiko (approved):
Looks pretty good, thanks!

PR #189, opened 2023-05-25
fix: sockfilter example output

JackyYin:
When we have multiple calls to `inet_ntoa()`, this function might return the same char* even though the underlying string has been modified.

We'll get the same string if we use this function in a row in a single printf function(because char* pointed to the same location).

On the other hand, support `-i <interface>` option in sockfilter program.

### DIFF
--- a/examples/c/sockfilter.c
+++ b/examples/c/sockfilter.c
@@ -1,5 +1,6 @@
 // SPDX-License-Identifier: GPL-2.0 OR BSD-3-Clause
 /* Copyright (c) 2022 Jacky Yin */
+#include <argp.h>
 #include <arpa/inet.h>
 #include <assert.h>
 #include <bpf/libbpf.h>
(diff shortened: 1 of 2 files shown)


examples/c/sockfilter.c
```diff
@@ -56,10 +56,17 @@ static int libbpf_print_fn(enum libbpf_print_level level, const char *format, va
 	return vfprintf(stderr, format, args);
 }
 
+static inline void ltoa(uint32_t addr, char *dst)
+{
+	snprintf(dst, 16, "%d.%d.%d.%d", (addr >> 24) & 0xFF, (addr >> 16) & 0xFF,
```
anakryiko, line 61: should this be %u?

examples/c/sockfilter.c
```diff
@@ -56,10 +56,17 @@ static int libbpf_print_fn(enum libbpf_print_level level, const char *format, va
...
+		 (addr >> 8) & 0xFF, (addr & 0xFF));
+}
+
 static int handle_event(void *ctx, void *data, size_t data_sz)
 {
 	const struct so_event *e = data;
 	char ifname[IF_NAMESIZE];
+	char sstr[16] = { 0 }, dstr[16] = { 0 };
```
anakryiko, line 69: nit: please use `= {}` instead of `= { 0 }`

examples/c/sockfilter.c
```diff
@@ -90,6 +98,21 @@ int main(int argc, char **argv)
 	struct ring_buffer *rb = NULL;
 	struct sockfilter_bpf *skel;
 	int err, prog_fd, sock;
+	int argp;
+	char *interface = "lo";
+
+	while ((argp = getopt(argc, argv, "hi:")) != -1) {
```
anakryiko, line 104: see bootstrap example which uses argp framework for passing arguments. Let's use just one way (argp).

examples/c/sockfilter.c
```diff
@@ -90,6 +98,21 @@ int main(int argc, char **argv)
 	struct ring_buffer *rb = NULL;
 	struct sockfilter_bpf *skel;
 	int err, prog_fd, sock;
+	int argp;
+	char *interface = "lo";
```
anakryiko, line 102: const char * ?

anakryiko (changes requested):
The changes make sense. But please split the fix and new argument into two separate commits, and let's consistently use argp?

examples/c/sockfilter.c
JackyYin: The range of each number is between 0 - 255, we can use `%u` to make sure no negative number will appear.

anakryiko (approved):
applied, but still not sure about strdup()

examples/c/sockfilter.c
```diff
@@ -15,6 +16,46 @@
...
+static const struct argp_option opts[] = {
+	{ "interface", 'i', "INTERFACE", 0, "Network interface to attach" },
+	{},
+};
+
+static error_t parse_arg(int key, char *arg, struct argp_state *state)
+{
+	switch (key) {
+	case 'i':
+		env.interface = arg;
```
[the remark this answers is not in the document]
JackyYin: The pointer of network interface we retrieved from `parse_arg` is the same as retrieve it from `char **argv`,
those pointers should not be released during program lifetime.

So I think it's OK to just use it without string duplication.

PR #233, opened 2023-10-31
add xmake android

guri87-byte:


guri87-byte:
Need to update the request.

PR #238, opened 2023-11-03
examples/c: Add task_iter

jordalgo:
This is a simple example of using BPF iterators, which iterates over all tasks on the host.

### DIFF
--- a/examples/c/task_iter.bpf.c
+++ b/examples/c/task_iter.bpf.c
@@ -1,0 +1,67 @@
+// SPDX-License-Identifier: GPL-2.0 OR BSD-3-Clause
+/* Copyright (c) 2023 Meta */
+#include "vmlinux.h"
+#include <bpf/bpf_helpers.h>
+#include <bpf/bpf_tracing.h>
+#include <bpf/bpf_core_read.h>
+#include "task_iter.h"
+
+char LICENSE[] SEC("license") = "Dual BSD/GPL";
+
+struct {
+	__uint(type, BPF_MAP_TYPE_PERCPU_ARRAY);
+	__uint(max_entries, 1);
+	__type(key, __u32);
+	__type(value, struct task_info);
+} task_info_buf SEC(".maps");
+
+struct task_struct___post514 {
+	unsigned int __state;
+} __attribute__((preserve_access_index));
+
+struct task_struct___pre514 {
+	long state;
+} __attribute__((preserve_access_index));
+
+static __u32 get_task_state(void *arg)
+{
+	if (bpf_core_field_exists(struct task_struct___pre514, state)) {
+		struct task_struct___pre514 *task = arg;
+		return task->state;
+	} else {
+		struct task_struct___post514 *task = arg;
+		return task->__state;
+	}
+}
+
+static __u32 zero = 0;
+
+SEC("iter/task")
+int get_tasks(struct bpf_iter__task *ctx)
+{
+	struct seq_file *seq = ctx->meta->seq;
+	struct task_struct *task = ctx->task;
+	struct task_info *t;
+	long res;
+
+	if (!task) {
+		return 0;
+	}
+
+	t = bpf_map_lookup_elem(&task_info_buf, &zero);
+	if (!t) {
+		return 0;
+	}
+
+	t->pid = task->tgid;
+	t->tid = task->pid;
+	t->state = get_task_state(task);
+
+	bpf_probe_read_kernel_str(t->comm, TASK_COMM_LEN, task->comm);
+
+	res = bpf_get_task_stack(task, t->kstack, sizeof(__u64) * MAX_STACK_LEN, 0);
+	t->kstack_len = res <= 0 ? res : res / sizeof(t->kstack[0]);
+
+	bpf_seq_write(seq, t, sizeof(struct task_info));
+	return 0;
+}
--- a/examples/c/task_iter.c
+++ b/examples/c/task_iter.c
@@ -1,0 +1,119 @@
+// SPDX-License-Identifier: (LGPL-2.1 OR BSD-2-Clause)
+/* Copyright (c) 2023 Meta */
+#include <argp.h>
+#include <signal.h>
+#include <stdio.h>
+#include <sys/resource.h>
+#include <bpf/bpf.h>
+#include <bpf/libbpf.h>
+#include <unistd.h>
+#include "task_iter.h"
+#include "task_iter.skel.h"
+
+static struct env {
+	bool verbose;
+} env;
+
+static int libbpf_print_fn(enum libbpf_print_level level, const char *format, va_list args)
+{
+	if (level == LIBBPF_DEBUG && !env.verbose)
+		return 0;
+	return vfprintf(stderr, format, args);
+}
+
+static volatile bool exiting = false;
+
+static void sig_handler(int sig)
+{
+	exiting = true;
+}
+
+static const char *get_task_state(__u32 state)
+{
+	// taken from
+	// https://elixir.bootlin.com/linux/latest/source/include/linux/sched.h#L85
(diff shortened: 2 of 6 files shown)


examples/c/task_iter.bpf.c
```diff
@@ -0,0 +1,68 @@
...
+
+struct task_struct___post514 {
+	unsigned int __state;
+} __attribute__((preserve_access_index));
+
+struct task_struct___pre514 {
+	long state;
+} __attribute__((preserve_access_index));
+
+__u32 get_task_state(void *arg)
```
anakryiko, line 26: make this static?

examples/c/task_iter.bpf.c
```diff
@@ -0,0 +1,68 @@
...
+{
+	if (bpf_core_field_exists(((struct task_struct___pre514 *)0)->state)) {
```
anakryiko, line 28: please use a more user-friendly form:

```
bpf_core_field_exists(struct task_struct___pre514, state);
```

examples/c/task_iter.bpf.c
```diff
@@ -0,0 +1,68 @@
...
+	}
+}
+
+SEC("iter/task")
+int get_tasks(struct bpf_iter__task *ctx)
+{
+	struct seq_file *seq = ctx->meta->seq;
+	struct task_struct *task = ctx->task;
+
+	if (task == (void *)0) { // end
```
anakryiko, line 43: let's not add C++-style comments to BPF code

and it's seems more appropriate to do `if (!task)` check, not sure why `(void *)0` was chosen

examples/c/task_iter.bpf.c
```diff
@@ -0,0 +1,68 @@
...
+		return 0;
+	}
+
+	__u32 zero = 0;
```
anakryiko, line 47: let's set a good example for this pattern, please use static global variable for zero.

examples/c/task_iter.bpf.c
```diff
@@ -0,0 +1,68 @@
...
+		return 0;
+	}
+
+	t->pid = task->tgid;
+	t->tid = task->pid;
+	t->state = get_task_state(task);
+
+	bpf_probe_read_kernel_str(t->comm, TASK_COMM_LEN, task->comm);
+
+	int64_t res = bpf_get_task_stack(task, t->kstack, sizeof(__u64) * MAX_STACK_LEN, 0);
```
anakryiko, line 59: we still follow C89 code style (kernel code style) for BPF code, so please move variable declarations to the top. And just use s64 instead of int64_t? you are already using __u32 and __u64 for task_iter struct, so seems fitting

examples/c/task_iter.bpf.c
```diff
@@ -0,0 +1,68 @@
...
+	if (res <= 0) {
+		t->kstack_len = 0;
+	} else {
+		t->kstack_len = res / sizeof(__u64);
+	}
```
anakryiko, line 64: maybe let's propagate error as well? something like

```
t->kstack_len = res < 0 ? res : res / sizeof(t->kstack[0]);
```

Note also t->kstack[0] for sizeof, it communicates better what we are doing here.

examples/c/task_iter.h
```diff
@@ -0,0 +1,16 @@
+/* SPDX-License-Identifier: (LGPL-2.1 OR BSD-2-Clause) */
+/* Copyright (c) 2020 Facebook */
+
+#define TASK_COMM_LEN 16
+#define MAX_STACK_LEN 127
+
+struct task_iter {
```
anakryiko, line 7: nit: maybe name it something more meaningful, like "task_info" or something? iter doesn't make much sense here

examples/c/task_iter.bpf.c
```diff
@@ -0,0 +1,68 @@
+// SPDX-License-Identifier: GPL-2.0 OR BSD-3-Clause
+/* Copyright (c) 2020 Facebook */
```
anakryiko, line 2: 2023? Meta?

examples/c/task_iter.c
```diff
@@ -0,0 +1,130 @@
+// SPDX-License-Identifier: (LGPL-2.1 OR BSD-2-Clause)
+/* Copyright (c) 2020 Facebook */
```
anakryiko, line 2: 2023? Meta?

examples/c/task_iter.c
```diff
@@ -0,0 +1,130 @@
...
+}
+
+static volatile bool exiting = false;
+
+static void sig_handler(int sig)
+{
+	exiting = true;
+}
+
+void print_task_state(__u32 state)
```
anakryiko, line 31: static

examples/c/task_iter.c
```diff
@@ -0,0 +1,130 @@
...
+
+	/* Load and verify BPF application */
+	skel = task_iter_bpf__open();
+	if (!skel) {
+		fprintf(stderr, "Failed to open and load BPF skeleton\n");
+		return 1;
+	}
+
+	/* Load & verify BPF programs */
+	err = task_iter_bpf__load(skel);
```
anakryiko, line 84: combine open and load with `task_iter_bpf__open_and_load()`?

examples/c/task_iter.c
```diff
@@ -0,0 +1,130 @@
...
+
+	/* Attach tracepoints */
+	err = task_iter_bpf__attach(skel);
+	if (err) {
+		fprintf(stderr, "Failed to attach BPF skeleton\n");
+		goto cleanup;
+	}
+
+	get_tasks_fd = bpf_iter_create(bpf_link__fd(skel->links.get_tasks));
+	if (!get_tasks_fd) {
```
anakryiko, line 98: this doesn't handle errors, should be `fd < 0`

examples/c/task_iter.c
```diff
@@ -0,0 +1,130 @@
...
+		err = -1;
+		fprintf(stderr, "Failed to create iter\n");
+		goto cleanup;
+	}
+
+	struct task_iter buf;
+	ssize_t ret;
```
anakryiko, line 105: let's move all variable declarations to the top of the function

examples/c/task_iter.c
```diff
@@ -0,0 +1,130 @@
...
+		printf("State: NOLOAD\n");
+		break;
+	case 0x0402:
+		printf("State: IDLE\n");
+		break;
+	case 0x0800:
+		printf("State: NEW\n");
+		break;
+	default:
+		printf("Unknown Task state\n");
```
anakryiko, line 59: lowercase task?

examples/c/task_iter.c
```diff
@@ -0,0 +1,130 @@
...
+		printf("State: WAKING\n");
+		break;
+	case 0x0400:
+		printf("State: NOLOAD\n");
+		break;
+	case 0x0402:
+		printf("State: IDLE\n");
+		break;
+	case 0x0800:
+		printf("State: NEW\n");
```
anakryiko, line 56: your code would be shorter and easier to use if this function returns `const char *` directly.

```
switch (state){
  case 0x0000: return "RUNNING";
  case 0x0001: return "INTERRUPTIBLE";
  ...
  default: return "<unknown>";
}
```

anakryiko (changes requested):

examples/c/task_iter.h
jordalgo: task_info works for me

examples/c/task_iter.c
jordalgo: oooo TIL

jordalgo:
@anakryiko Thanks for the review. I think I made all the suggested changes.

anakryiko (approved):
You still have C++ comments in some places, but I'll fix it up locally. Thanks for a useful example!

Issue #243, opened 2023-12-21
example uprobe no output 


zaiqiang-dong:
build example/uprobe and run success. but  no output when sudo cat /sys/kernel/debug/tracing/trace_pipe

root@NB:/home/zaiqdong/work/bin# ./uprobe
libbpf: loading object 'uprobe_bpf' from buffer
libbpf: elf: section(2) uprobe, size 64, link 0, flags 6, type=1
libbpf: sec 'uprobe': found program 'uprobe_add' at insn offset 0 (0 bytes), code size 8 insns (64 bytes)
libbpf: elf: section(3) uretprobe, size 56, link 0, flags 6, type=1
libbpf: sec 'uretprobe': found program 'uretprobe_add' at insn offset 0 (0 bytes), code size 7 insns (56 bytes)
libbpf: elf: section(4) uprobe//proc/self/exe:uprobed_sub, size 64, link 0, flags 6, type=1
libbpf: sec 'uprobe//proc/self/exe:uprobed_sub': found program 'uprobe_sub' at insn offset 0 (0 bytes), code size 8 insns (64 bytes)
libbpf: elf: section(5) uretprobe//proc/self/exe:uprobed_sub, size 56, link 0, flags 6, type=1
libbpf: sec 'uretprobe//proc/self/exe:uprobed_sub': found program 'uretprobe_sub' at insn offset 0 (0 bytes), code size 7 insns (56 bytes)
libbpf: elf: section(6) license, size 13, link 0, flags 3, type=1
libbpf: license of uprobe_bpf is Dual BSD/GPL
libbpf: elf: section(7) .rodata, size 128, link 0, flags 2, type=1
libbpf: elf: section(8) .BTF, size 1599, link 0, flags 0, type=1
libbpf: elf: section(9) .BTF.ext, size 328, link 0, flags 0, type=1
libbpf: elf: section(10) .symtab, size 360, link 18, flags 0, type=2
libbpf: elf: section(11) .reluprobe, size 16, link 10, flags 0, type=9
libbpf: elf: section(12) .reluretprobe, size 16, link 10, flags 0, type=9
libbpf: elf: section(13) .reluprobe//proc/self/exe:uprobed_sub, size 16, link 10, flags 0, type=9
libbpf: elf: section(14) .reluretprobe//proc/self/exe:uprobed_sub, size 16, link 10, flags 0, type=9
libbpf: looking for externs among 15 symbols...
libbpf: collected 0 externs total
[...]

PR #244, opened 2024-01-03
examples/c: prevent uprobe_add/uprobe_sub inlining

anakryiko:
Depending on compiler, uprobe_add/uprobe_sub can still be inlined. Use no-op volatile asm and noinline attribute to prevent compiler from inlining. Just global function is not enough.

### DIFF
--- a/examples/c/uprobe.c
+++ b/examples/c/uprobe.c
@@ -12,14 +12,19 @@ static int libbpf_print_fn(enum libbpf_print_level level, const char *format, va
 	return vfprintf(stderr, format, args);
 }
 
-/* It's a global function to make sure compiler doesn't inline it. */
-int uprobed_add(int a, int b)
+/* It's a global function to make sure compiler doesn't inline it. To be extra
+ * sure we also use "asm volatile" and noinline attributes to prevent
+ * compiler from local inlining.
+ */
+__attribute__((noinline)) int uprobed_add(int a, int b)
 {
+	asm volatile ("");
 	return a + b;
 }
 
-int uprobed_sub(int a, int b)
+__attribute__((noinline)) int uprobed_sub(int a, int b)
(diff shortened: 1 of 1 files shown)


examples/c/uprobe.c
```diff
@@ -12,14 +12,19 @@ static int libbpf_print_fn(enum libbpf_print_level level, const char *format, va
 	return vfprintf(stderr, format, args);
 }
 
-/* It's a global function to make sure compiler doesn't inline it. */
-int uprobed_add(int a, int b)
+/* It's a global function to make sure compiler doesn't inline it. To be extra
+ * sure we also use "asm volatile" and noinline attributes to prevent
+ * compiler from local inlining.
```
danielocfb, line 17: What do you mean by local inlining?

danielocfb (approved):
Looks good to me.

examples/c/uprobe.c
anakryiko: compiler can choose to inline calls to uprobe_add() function from other functions defined in the same .c file, but still keep uprobe_add() around (because it's global, so could be, technically, called from other .c/.o files). This is probably what's happening here, as uprobe_add seems to be attachable (we get no error finding its offset), but it's not really called at runtime, because all the calls are actually inlined.

For global functions, adding `__attribute__(weak)` should help and be enough. In general, for static functions it's way more problematic. You need `__attribute__((noinline))` and `asm volatile ();` trick (I don't think just one of those is enough). So in this case I think just having weak would be ok, but noinline+asm volatile is kind of universal "big hammer" approach I went with.

anakryiko: oh, in case you haven't seen, this was in response to https://github.com/libbpf/libbpf-bootstrap/issues/243

PR #248, opened 2024-01-12
examples/rust: Store skeletons in $OUT_DIR

danielocfb:
Store skeletons in $OUT_DIR as opposed to .output, which is kind of the proper thing to do (though it certainly doesn't help discoverability...) and it is more in line with how examples in libbpf-rs work. Also remove the "there is hope!" comments, because even with https://github.com/rust-lang/rust/pull/83366 resolved we still cannot use the concat! macro in #[path = ...] attributes. All hope is lost.

### DIFF
--- a/examples/rust/profile/build.rs
+++ b/examples/rust/profile/build.rs
@@ -1,4 +1,6 @@
 use std::env;
+use std::env::consts::ARCH;
+use std::path::Path;
 use std::path::PathBuf;
 
 use libbpf_cargo::SkeletonBuilder;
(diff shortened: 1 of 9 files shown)


anakryiko (approved):
Do you think it's possible to support something like that in libbpf-cargo itself, so that users don't have to duplicate this logic all the time?

Either way, landing as it is an improvement that fixes a real problem, thanks!

danielocfb:
> Do you think it's possible to support something like that in libbpf-cargo itself, so that users don't have to duplicate this logic all the time?

I thought about it, but it's not clear to me how that would be possible in a generic fashion. It would probably mean that we'd have to ship the headers ourselves, which is ugly. The cleanest solution would be to unconditionally extract the header from BTF, which has its own downsides. It's going to be a mess either way, it seems.

anakryiko:
> I thought about it, but it's not clear to me how that would be possible in a generic fashion. It would probably mean that we'd have to ship the headers ourselves, which is ugly. The cleanest solution would be to unconditionally extract the header from BTF, which has its own downsides. It's going to be a mess either way, it seems.

All that seems like an overkill, agreed. I think for libbpf-bootstrap we should just create directory symlinks to match Rust's ARCH values, and avoid those matches.

Issue #250, opened 2024-01-23
[Android] Failed to find valid kernel BTF


zebede22:
Device: Android 13 AOSP for x86_64, all necessary flags are enabled during kernel build, including `CONFIG_DEBUG_INFO_BTF`

bpftool version: `not sure, provided by xmake-repo`
libbpf: `432a4348579e0049af45c6712a2c1430f607b086`

```bash
> file /sys/kernel/btf/vmlinux
/sys/kernel/btf/vmlinux : data
```

Steps:
```bash
git clone --recurse-submodules https://github.com/libbpf/libbpf-bootstrap
...
xmake f -p android -a x86_64 -c --require-bpftool=y
xmake -y
```
all compiles successfully.

`minimal` works on the devices.
`bootstrap` fails:
```bash
libbpf: failed to find valid kernel BTF
libbpf: Error loading vmlinux BTF: -3
libbpf: failed to load object 'bootstrap_bpf'
libbpf: failed to load BPF skeleton 'bootstrap_bpf': -3
Failed to load and verify BPF skeleton
```

My guess is that the problem is related to maps since minimal doesn't contain any and it works.

zebede22:
I think @waruqi  is able to help here

waruqi:
I don't have an android environment to run and test it, you can run `xmake -rvD` to check the full compilation flags.

zebede22:
Here is flags but not sure how it helps:
`
[...]

PR #253, opened 2024-01-29
libbpf: update to latest upstream version

anakryiko:
Pull in latest libbpf upstream version that contains a fix for file access checks on Android ([0]).

  [0] 830e0d017bb1 ("libbpf: Fix faccessat() usage on Android"))

Closes: https://github.com/libbpf/libbpf-bootstrap/issues/250


anakryiko:
@waruqi, @slain01 latest libbpf is using `memfd_create()` libc API, which I thought it present in Android, but apparently not on all versions. Given you guys work with Android, can you please help me figure out best way to deal with it? Is there some way to detect if it's present/which version/etc? Or it would be enough to just specify minimum Android version, etc?

waruqi:
> @waruqi, @slain01 latest libbpf is using `memfd_create()` libc API, which I thought it present in Android, but apparently not on all versions. Given you guys work with Android, can you please help me figure out best way to deal with it? Is there some way to detect if it's present/which version/etc? Or it would be enough to just specify minimum Android version, etc?

we can try to use weak symbol to detect it dynamically

```c
extern __attribute((weak)) int memfd_create(const char* _Nonnull __name, unsigned __flags);
if (memfd_create) {
    // TODO
}
```

anakryiko:
> we can try to use weak symbol to detect it dynamically
> 
> ```c
> extern __attribute((weak)) int memfd_create(const char* _Nonnull __name, unsigned __flags);
> if (memfd_create) {
>     // TODO
> }
> ```

I ended up going a slightly other direction, but it seems to work (at least it compiles). See https://github.com/libbpf/libbpf/commit/4b3f89a68cb9db314aaf973a5f4f363386b12839

If someone with access to Android can confirm that it actually works well at runtime, I'd really appreciate it. Meanwhile I'm going to submit another patch upstream and re-sync it back into Github.

anakryiko:
Ok, seems like we are good to go.

PR #259, opened 2024-03-08
ci: run build workflows daily

anakryiko:
Let's have regular daily CI build test runs.

### DIFF
--- a/.github/workflows/build-android.yml
+++ b/.github/workflows/build-android.yml
@@ -5,6 +5,8 @@ on:
     branches:
       - master
   pull_request:
+  schedule:
+    - cron: '7 17 * * *'
 
 jobs:
   build_libbpf_bootstrap_android:
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -5,6 +5,8 @@ on:
     branches:
       - master
   pull_request:
+  schedule:
+    - cron: '7 17 * * *'
 
 jobs:
   build_libbpf_bootstrap:


anakryiko:
@waruqi it seems like Android builds started failing (and we didn't have daily CI runs to notice in time, oops). Can you please check what's wrong and help fixing. It's failing with:

```
  => install linux-tools 5.16.9 .. failed
```

But no details are available in CI.

waruqi:
It seems that after upgrading ci's linux environment, bpftool fails to execute when building linux tools. I'm not sure why at the moment, but maybe I need to update the linux kernel version, and it may take some time to fix it.

waruqi:
I have fixed it.

anakryiko:
> I have fixed it.

great, thank you! for history, do you have a link to the fix, if anyone would be curious to check?

waruqi:
I have updated linux tools (kernel) to 6.x, see 
https://github.com/xmake-io/xmake-repo/pull/3371

becasue linux action ci has been upgraded

anakryiko:
> After compiling bpftool, make calls bpftool to dump vmlinux. calling 5.x kernel/bpftool on a newer linux kernel fails, I don't know why, maybe it's strongly correlated to the linux kernel version.

newer kernel has newer BTF (e.g., types like ENUM64), which old bpftool won't recognize and will refuse to work with BTF. This is probably what you are running into.

anakryiko:
@waruqi I'm not sure why, but now Android build tests are timing out after 6 hours, being stuck. See https://github.com/libbpf/libbpf-bootstrap/actions/runs/8253131606/job/22586521612 as an example. One of the runs did succeed (https://github.com/libbpf/libbpf-bootstrap/actions/runs/8253131606/job/22586521503), so it's not 100% rate, but I don't think we had a successful full run just yet.

waruqi:
I will look at it.

waruqi:
> @waruqi I'm not sure why, but now Android build tests are timing out after 6 hours, being stuck. See https://github.com/libbpf/libbpf-bootstrap/actions/runs/8253131606/job/22586521612 as an example. One of the runs did succeed (https://github.com/libbpf/libbpf-bootstrap/actions/runs/8253131606/job/22586521503), so it's not 100% rate, but I don't think we had a successful full run just yet.

I've fixed this, but it requires the master version to work now, I'll update it to latest after the next release.

https://github.com/libbpf/libbpf-bootstrap/pull/263

Issue #257, opened 2024-03-02
when compile the Rust example in libbpf-bootstrap for target  arm64 ，why build.rs use vmlinux.h is x86 ?


lichm1007:
eg: xdp -> build.rs  
when use command  :   RUSTFLAGS="-C linker=/usr/bin/aarch64-linux-gnu-gcc" cargo build --target aarch64-unknown-linux-gnu   
the build.rs -> ARCH is x86_64 
 
**fn main() {
    let mut out =
        PathBuf::from(env::var_os("OUT_DIR").expect("OUT_DIR must be set in build script"));
    out.push("xdppass.skel.rs");

    ### **println!("arch={}", ARCH);   // is  x86_64   ?**  

    SkeletonBuilder::new()
        .source(SRC)
        .clang_args(format!(
            "-I{}",
            Path::new("../../../vmlinux")
                .join(match ARCH {
                    "aarch64" => "arm64",
                    "loongarch64" => "loongarch",
                    "powerpc64" => "powerpc",
                    "riscv64" => "riscv",
                    "x86_64" => "x86",
                    _ => ARCH,
                })
                .display()
        ))
        .build_and_generate(out)
        .unwrap();
    println!("cargo:rerun-if-changed={}", SRC);
}**

danielocfb:
It's conceivable that `ARCH` does not honor the cross compilation target. In that case we may actually have to use the `CARGO_CFG_TARGET_ARCH` environment variable. Could you try that out, by any chance?

lichm1007:
use command :  CARGO_CFG_TARGET_ARCH=aarch64 RUSTFLAGS="-C linker=/usr/bin/aarch64-linux-gnu-gcc" cargo build --target aarch64-unknown-linux-gnu   ，  build.rs  arch is x86_64   ， can not set to aarch64

danielocfb:
What I meant is:
```patch
--- examples/rust/xdp/build.rs
+++ examples/rust/xdp/build.rs
@@ -1,5 +1,4 @@
 use std::env;
-use std::env::consts::ARCH;
 use std::path::Path;
 use std::path::PathBuf;

@@ -12,18 +11,21 @@ fn main() {
         PathBuf::from(env::var_os("OUT_DIR").expect("OUT_DIR must be set in build script"));
     out.push("xdppass.skel.rs");

+    let arch = env::var("CARGO_CFG_TARGET_ARCH")
[...]

PR #261, opened 2024-03-11
examples/rust: Use `CARGO_CFG_TARGET_ARCH` environment variable

danielocfb:
As per #257, it appears as if `std::env::consts::ARCH` does not get updated to the target architecture in a cross-compilation context. Switch to using the `CARGO_CFG_TARGET_ARCH` environment variable in the hopes that we get the desired behavior.


anakryiko:
Great, thanks for fixing!

PR #258, opened 2024-03-08
fix missing blazesym head file

hzhuang1:
There're some updates in Makefile in commit 4b207312dd. But it's not synchronized into CMakeLists.txt. It causes the error in below.

 77%] Building C object CMakeFiles/profile.dir/profile.c.o
libbpf-bootstrap/examples/c/profile.c:17:10: fatal error: blazesym.h: No such file or directory
   17 | #include "blazesym.h"
      |          ^~~~~~~~~~~~
compilation terminated.
make[2]: *** [CMakeFiles/profile.dir/build.make:84: CMakeFiles/profile.dir/profile.c.o] Error 1
make[1]: *** [CMakeFiles/Makefile2:520: CMakeFiles/profile.dir/all] Error 2
make: *** [Makefile:91: all] Error 2

PR #262, opened 2024-03-11
examples/c: Use CMake stuff in CI

danielocfb:
As #258 showed, we also have a CMake based build infrastructure lying around for the examples. However, currently it is not used in CI and, hence, prone to breakage.
Change that fact by building using CMake as part of the build workflow.


anakryiko:
Thanks, we now have all three build systems covered, yay.

PR #280, opened 2024-07-14
cmake: Allow additional header deps for `bpf_object`

ShawnZhong:
It is quite common that a `.bpf.c` file depends on a header file. 

However, the current `FindBpfObject.cmake` does not allow one to specify such dependency, so modification to the header file does not trigger recompilation of the BPF object file (and subsequent skeleton header).

This PR addresses the issue by allowing one to write, for example, 

```
bpf_object(app app.bpf.c app.h utils.h)
```

So that the change to the header files (`app.h` and `utils.h` is tracked).

### DIFF
--- a/tools/cmake/FindBpfObject.cmake
+++ b/tools/cmake/FindBpfObject.cmake
@@ -41,18 +41,19 @@ as the associated dependencies.
 
 .. code-block:: cmake
 
-  bpf_object(<name> <source>)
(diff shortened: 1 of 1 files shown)


ShawnZhong:
cc: @danobi who wrote the original `FindBpfObject.cmake`

danobi:
sorry, missed notification. LGTM

PR #289, opened 2024-10-03
Replace `libbpf` submodule with symbolic link in `bpftool`

yunwei37:
This PR removes the `libbpf` submodule in favor of a symbolic link to the `libbpf` directory within `bpftool`. The goal is to reduce redundancy and make dependency management easier, as both projects share the same `libbpf` implementation.

#### Key Changes:

- Removed the `libbpf` submodule from the repository.
- Created a symbolic link to `bpftool`'s `libbpf` directory.
  
#### Benefits:

- **Consistency**: Both `libbpf` instances are always in sync between `bpftool` and this repository.
- **Simplified Workflow**: No need to update the `libbpf` submodule separately.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,6 +1,3 @@
-[submodule "libbpf"]
-	path = libbpf
-	url = https://github.com/libbpf/libbpf.git
 [submodule "bpftool"]
 	path = bpftool
 	url = https://github.com/libbpf/bpftool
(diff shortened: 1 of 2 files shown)


anakryiko:
I don't think it's appropriate to do. bpftool's libbpf is bpftool's internal implementation detail, I don't see the need to assume it. Does this extra submodule cause any problems in practice?

yunwei37:
OK, I see. Thanks! I will close that.

Just some people complain about the extra submodule they need to download when their network is slow.

PR #294, opened 2024-11-03
Remove backing qdisc on cleanup

pcolladosoto:
Hi! I've been playing around with Go's bindings for `libbpf` (i.e. [`aquasecurity/libbpfgo`](https://github.com/aquasecurity/libbpfgo)) and found that their example of how to use BPF-based TC filters did not clean the backing `qdisc` after the cleanup.

Even though this behaviour was not causing any issues, I decided to dive into `libbpf`'s implementation and found that, indeed, a backing `qdisc` is only removed if the attach point of the hook is `BPF_TC_INGRESS|BPF_TC_EGRESS` which is usually not the case. In that way, `libbpf`'s API is 'forcing' the user to explicitly request the clearing of the `qdisc`. This behaviour is documented in the associated [patch series](https://patchwork.kernel.org/project/netdevbpf/patch/20210428162553.719588-3-memxor@gmail.com/).

Anyway, this PR simply forces the removal of the backing `qdisc` by setting `tc_hook.attach_point = BPF_TC_INGRESS|BPF_TC_EGRESS;` just before the call to `bpf_tc_hook_destroy`. Given the example is rather self-contained I suppose that makes sense? The rationale behind opening up the PR is that it took me several hours to get to the patch series and I was hoping the next person wouldn't have to do the same... Maybe this can be left as a comment in `tc.c` even if one decides not to remove the `qdisc`?

In a real-world approach I suppose one could use the return value of `bpf_tc_hook_create` to decide whether to remove the `qdisc` or not so that the backing `qdisc` is only removed if the return value is `-EEXIST`? Judging by the comment above the call to `bpf_tc_hook_create` I suppose this behaviour was known by the original author of `tc.c`, but I feel like a clarification would be great for people using these examples as scaffolding for larger projects.

Thanks a ton for all this work! It's really helped me up to this moment and I'm sure it'll continue to do so for quite a while!

### DIFF
--- a/examples/c/tc.c
+++ b/examples/c/tc.c
@@ -78,8 +78,13 @@ int main(int argc, char **argv)
 	}
 
 cleanup:
-	if (hook_created)
+	if (hook_created) {
+		/* Explicitly ask for the backing qdisc to be destroyed. Otherwise, only the BPF filters will be removed.
+		 * See https://patchwork.kernel.org/project/netdevbpf/patch/20210428162553.719588-3-memxor@gmail.com/
+		 */
+		tc_hook.attach_point = BPF_TC_INGRESS|BPF_TC_EGRESS;
 		bpf_tc_hook_destroy(&tc_hook);
+	}
 	tc_bpf__destroy(skel);
 	return -err;
 }


pcolladosoto:
Hi @kkdwivedi! Thanks a ton for the clarification. After going through the references on LWN I see how the `netlink` API is indeed not flexible enough to accommodate this user case in a straightforward way. I'll look into migrating to the new `tcx` API for my project.

Given my current use case I can safely assume my program 'owns' the qdisc I attach the eBPF program to in the sense that no other processes make use of it (i.e. I'm not working in the context of virtualisation in any way, shape or form).

However, as you correctly point out this might not (and certainly won't) be the case and anybody wanting to remove the qdisc should want it enough to do all this digging. With that in mind, maybe adding a bit of documentation (maybe in the shape of a comment?) would be a compromise? That way there's no qdisc deletion at all but users are pointed both in the right direction for deleting the qdisc and maybe also the new `tcx` API.

I completely understand that you might just prefer to leave things as they are. In that case feel free to close the PR.

Anyway, thanks a ton for the reference to the patch series: I learned a lot thanks to it!

All the best :)

PR #319, opened 2024-12-05
Profile: command line switch for virtual machine

### LINKED ISSUE #313: Optimize profile demo for virtual machine compatibility
When running the `profile` demo on a virtual machine, I encountered the following error:
```
# C
sudo ./profile 
Fail to set up performance monitor on a CPU/Core

# Rust
sudo ./target/release/profile
Error: Error: failed to open perf event

Caused by:
    No such file or directory (os error 2)
```
Thanks to https://github.com/libbpf/libbpf-bootstrap/issues/210#issuecomment-1698392516, it seems that in the virtual machine the `profile` demo needs different attributes to work as expected. 

Patches on virtual machines can be:

For C:
```diff
diff --git a/examples/c/profile.c b/examples/c/profile.c
index 1eab837..23c1e1e 100644
--- a/examples/c/profile.c
+++ b/examples/c/profile.c
@@ -206,9 +206,9 @@ int main(int argc, char *const argv[])
        links = calloc(num_cpus, sizeof(struct bpf_link *));

        memset(&attr, 0, sizeof(attr));
-       attr.type = PERF_TYPE_HARDWARE;
+       attr.type = PERF_TYPE_SOFTWARE;
        attr.size = sizeof(attr);
-       attr.config = PERF_COUNT_HW_CPU_CYCLES;
+       attr.config = PERF_COUNT_SW_CPU_CLOCK;
        attr.sample_freq = freq;
        attr.freq = 1;
```

For Rust:
```diff
diff --git a/examples/rust/profile/src/main.rs b/examples/rust/profile/src/main.rs
index de03e41..dd65017 100644
--- a/examples/rust/profile/src/main.rs
+++ b/examples/rust/profile/src/main.rs
@@ -52,9 +52,9 @@ fn init_perf_monitor(freq: u64) -> Result<Vec<i32>, libbpf_rs::Error> {
             buf.leak().as_mut_ptr() as *mut syscall::perf_event_attr
         )
     };
-    attr._type = syscall::PERF_TYPE_HARDWARE;
+    attr._type = syscall::PERF_TYPE_SOFTWARE;
     attr.size = mem::size_of::<syscall::perf_event_attr>() as u32;
-    attr.config = syscall::PERF_COUNT_HW_CPU_CYCLES;
+    attr.config = syscall::PERF_COUNT_SW_CPU_CLOCK;
     attr.sample.sample_freq = freq;
     attr.flags = 1 << 10; // freq = 1
     (0..nprocs)
```
```diff
diff --git a/examples/rust/profile/src/syscall.rs b/examples/rust/profile/src/syscall.rs
index eb73f3b..e697ded 100644
--- a/examples/rust/profile/src/syscall.rs
+++ b/examples/rust/profile/src/syscall.rs
@@ -54,7 +54,9 @@ pub struct perf_event_attr {
 }

 pub const PERF_TYPE_HARDWARE: u32 = 0;
+pub const PERF_TYPE_SOFTWARE: u32 = 1;
 pub const PERF_COUNT_HW_CPU_CYCLES: u64 = 0;
+pub const PERF_COUNT_SW_CPU_CLOCK: u64 = 0;

 extern "C" {
     fn syscall(number: libc::c_long, ...) -> libc::c_long;
```

It will be nice the `profile` demo works on both physical machines and virtual machines.

- use correct attributes based on detect whether the profile demo is running on a virtual machine or a physical machine at runtime

OR
- add a note in the profile demo’s section of the README

flearc:
Fix #313 

This PR introduces a new command line argument to the profile example, allowing users to specify the platform type (`host` or `guest`, default is `host`). This ensures the correct performance monitoring attributes are used based on user's input.

Key Changes:
1. Adjust performance monitoring attributes based on the `-p` argument.
2. Provide a hint when a `perf_event_open` error occurs, recommending the `-p guest` option for virtual machines.

### DIFF
--- a/examples/c/profile.c
+++ b/examples/c/profile.c
@@ -3,6 +3,7 @@
 #include <assert.h>
 #include <stdint.h>
 #include <stdio.h>
+#include <getopt.h> 
 #include <unistd.h>
 #include <string.h>
 #include <fcntl.h>
@@ -133,13 +134,17 @@ static int event_handler(void *_ctx, void *data, size_t size)
 
 static void show_help(const char *progname)
 {
-	printf("Usage: %s [-f <frequency>] [-h]\n", progname);
+	printf("Usage: %s [-f <frequency>] [--sw-event] [-h]\n", progname);
+	printf("Options:\n");
+	printf("  -f <frequency>  Sampling frequency [default: 1]\n");
+	printf("  --sw-event      Use software event for triggering stack trace capture\n");
+	printf("  -h              Print help\n");
 }
 
 int main(int argc, char *const argv[])
 {
 	const char *online_cpus_file = "/sys/devices/system/cpu/online";
-	int freq = 1, pid = -1, cpu;
+	int freq = 1, sw_event = 0, pid = -1, cpu;
 	struct profile_bpf *skel = NULL;
 	struct perf_event_attr attr;
 	struct bpf_link **links = NULL;
@@ -149,12 +154,20 @@ int main(int argc, char *const argv[])
 	int argp, i, err = 0;
 	bool *online_mask = NULL;
 
-	while ((argp = getopt(argc, argv, "hf:")) != -1) {
+	static struct option long_options[] = {
+		{"sw-event", no_argument, 0, 's'},
+		{0, 0, 0, 0}
+	};
+
+	while ((argp = getopt_long(argc, argv, "hf:", long_options, NULL)) != -1) {
 		switch (argp) {
 		case 'f':
 			freq = atoi(optarg);
 			if (freq < 1)
 				freq = 1;
+			break;
+		case 's':
+			sw_event = 1;
 			break;
 
 		case 'h':
@@ -206,9 +219,9 @@ int main(int argc, char *const argv[])
 	links = calloc(num_cpus, sizeof(struct bpf_link *));
 
 	memset(&attr, 0, sizeof(attr));
-	attr.type = PERF_TYPE_HARDWARE;
+	attr.type = sw_event ? PERF_TYPE_SOFTWARE : PERF_TYPE_HARDWARE;
 	attr.size = sizeof(attr);
-	attr.config = PERF_COUNT_HW_CPU_CYCLES;
+	attr.config = sw_event ? PERF_COUNT_SW_CPU_CLOCK : PERF_COUNT_HW_CPU_CYCLES;
 	attr.sample_freq = freq;
 	attr.freq = 1;
 
@@ -220,7 +233,13 @@ int main(int argc, char *const argv[])
 		/* Set up performance monitoring on a CPU/Core */
 		pefd = perf_event_open(&attr, pid, cpu, -1, PERF_FLAG_FD_CLOEXEC);
 		if (pefd < 0) {
-			fprintf(stderr, "Fail to set up performance monitor on a CPU/Core\n");
+			if (!sw_event && errno == ENOENT) {
+				fprintf(stderr,
+					"Fail to set up performance monitor on a CPU/Core.\n"
+					"Try running the profile example with the `--sw-event` option.\n");
+			} else {
+				fprintf(stderr, "Fail to set up performance monitor on a CPU/Core.\n");
+			}
 			err = -1;
 			goto cleanup;
 		}
--- a/examples/rust/profile/src/main.rs
+++ b/examples/rust/profile/src/main.rs
@@ -43,7 +43,7 @@ struct stacktrace_event {
     ustack: [u64; MAX_STACK_DEPTH],
 }
 
-fn init_perf_monitor(freq: u64) -> Result<Vec<i32>, libbpf_rs::Error> {
+fn init_perf_monitor(freq: u64, sw_event: bool) -> Result<Vec<i32>, libbpf_rs::Error> {
     let nprocs = libbpf_rs::num_possible_cpus().unwrap();
     let pid = -1;
     let buf: Vec<u8> = vec![0; mem::size_of::<syscall::perf_event_attr>()];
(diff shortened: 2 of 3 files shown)


danielocfb (commented):
Thanks for the pull request. My suggestion would be to not talk about virtual machines at all. A virtual machine may very well support the physical thing and vice versa. So I think we should instead talk about software event or something like that. Basically, call the switch `--sw-event` or something along those lines. Also, we may want to hint at its presence when `perf_event_open` fails with `ENOENT` or whatever the error.

flearc:
> Thanks for the pull request. My suggestion would be to not talk about virtual machines at all. A virtual machine may very well support the physical thing and vice versa. So I think we should instead talk about software event or something like that. Basically, call the switch `--sw-event` or something along those lines. Also, we may want to hint at its presence when `perf_event_open` fails with `ENOENT` or whatever the error.
> 
> Edit: Sorry, you already did the latter partly. But let's not do it on generic failure but for specific error.

Thanks for your suggestion! A command line option for more specific event is much better.
I have updated the PR and now only `ENOENT` error will print software event hint for user.

examples/c/profile.c
```diff
@@ -220,7 +229,13 @@ int main(int argc, char *const argv[])
 		/* Set up performance monitoring on a CPU/Core */
 		pefd = perf_event_open(&attr, pid, cpu, -1, PERF_FLAG_FD_CLOEXEC);
 		if (pefd < 0) {
-			fprintf(stderr, "Fail to set up performance monitor on a CPU/Core\n");
+			if (errno == ENOENT) {
```
danielocfb, line 232: Probably confusing if we fail to set up the software event but recommend that as a work around:
```suggestion
			if (errno == ENOENT && !sw_event) {
```

examples/rust/profile/src/main.rs
```diff
@@ -52,17 +52,32 @@ fn init_perf_monitor(freq: u64) -> Result<Vec<i32>, libbpf_rs::Error> {
...
     attr.flags = 1 << 10; // freq = 1
     (0..nprocs)
         .map(|cpu| {
             let fd = syscall::perf_event_open(attr.as_ref(), pid, cpu as i32, -1, 0) as i32;
             if fd == -1 {
-                Err(libbpf_rs::Error::from(io::Error::last_os_error()))
-                    .context("failed to open perf event")
+                let os_error = io::Error::last_os_error();
+                let error_context = match os_error.kind() {
+                    io::ErrorKind::NotFound => {
```
danielocfb, line 74: Similar comment as for C code above.

examples/rust/profile/src/main.rs
```diff
@@ -224,6 +239,9 @@ struct Args {
     /// Increase verbosity (can be supplied multiple times).
     #[arg(short = 'v', long = "verbose", global = true, action = ArgAction::Count)]
     verbosity: u8,
+    /// Software event
```
danielocfb, line 242: May be worth it to elaborate a bit more. Perhaps something like:
```suggestion
    /// Use software event for triggering stack trace capture.
    ///
    /// This can be useful for compatibility reasons if the hardware event is not available
    /// (which could happen in a virtual machine, for example).
```

examples/rust/profile/src/main.rs
```diff
@@ -224,6 +239,9 @@ struct Args {
...
+    #[arg(long = "sw-event", action = ArgAction::SetTrue)]
```
danielocfb, line 243: Haven't tried it, but my understanding is that the default action for `bool`s is to set it to true, so you can probably omit the `action = ...` part here.

examples/c/profile.c
```diff
@@ -149,13 +150,21 @@ int main(int argc, char *const argv[])
...
+	};
+
+	while ((argp = getopt_long(argc, argv, "hf:", long_options, NULL)) != -1) {
 		switch (argp) {
 		case 'f':
 			freq = atoi(optarg);
 			if (freq < 1)
 				freq = 1;
 			break;
+		case 1:
```
danielocfb, line 165: Nit: This `1` seems pretty magic-y. I had to page in `getopt_long` behavior. Perhaps we could use `'s'` or something instead of `1`, and/or introduce a constant of sorts?

danielocfb (approved):
Looks great to me, thanks! Left a few more minor comments. Can you please squash everything into a single commit (or two, for C & Rust, respectively, if you want) and we can go ahead with the merge.

examples/rust/profile/src/main.rs
flearc: Tried. The action part indeed don't need here, it's the default behavior.

examples/rust/profile/src/main.rs
flearc: Completed the help info. Also add it in c version.

examples/c/profile.c
flearc: You are right, only need to print hint without `--sw-event`.

examples/c/profile.c
flearc: I changed `1` to `s`.

examples/rust/profile/src/main.rs
flearc: Changed it and now use if statement like this.
```rust
let mut error_context = "Failed to open perf event.";
let os_error = io::Error::last_os_error();
if !sw_event && os_error.kind() == io::ErrorKind::NotFound {
        error_context = "Failed to open perf event.\n\
                         Try running the profile example with the `--sw-event` option.";
}
```

flearc:
@danielocfb Thanks for your helpful review. Just fixed these problems and squashed into one commit.

examples/rust/profile/src/main.rs
flearc: In c, the help info now:
```text
Usage: ./profile [-f <frequency>] [--sw-event] [-h]
Options:
  -f <frequency>  Sampling frequency [default: 1]
  --sw-event      Use software event for triggering stack trace capture
  -h              Print help
```

danielocfb (approved):
Looks great to me, thanks!

PR #343, opened 2025-06-02
Don't specify minimum Rust version for examples

d-e-s-o:
It's not actually clear that specifying the minimum supported Rust version is something we should do in these examples. Dependencies should carry their own requirements and these examples by now should build on the majority of Rust versions out there, as they don't use any advanced features and `1.71`, even assuming that it is accurate [which it may not be], is fairly "old" at this point.
As such, the specified minimum Rust version is more of a maintenance burden, as it can clash with *higher* requirements caused by dependencies (see #342 as an example). Remove it.


danielocfb:
Failures seem to be due to the daily Ubuntu woes and are unrelated.

PR #367, opened 2026-05-04
examples/c: Fix ringbuf leaks in  C libbpf examples

vdasu:
Three libbpf-bootstrap examples (`bootstrap`, `profile`, and `sockfilter`) call `bpf_ringbuf_reserve()` to obtain an event buffer, populate a subset of the event's fields, and emit the full `sizeof(*event)` bytes via `bpf_ringbuf_submit()`. `bpf_ringbuf_reserve()` does not zero-initialize the returned memory. Any bytes the source-level path leaves unwritten retain whatever the slot held previously. This can leak previously emitted record content back to userspace on subsequent events.

The `bootstrap` leak was tested on Linux 6.8. `struct event` is written by two asymmetric handlers (`handle_exec` and `handle_exit`), each of which leaves the other handler's fields untouched, plus a 4-byte padding hole that neither handler writes. Across 3,000 captured events, exit events leaked prior exec records' `filename` paths and exec events leaked prior exit records' `exit_code`/`duration_ns` bytes.

 `__builtin_memset(event, 0, sizeof(*event))` does not work for `profile` since the struct (2080 bytes) exceeds LLVM's inline-store budget. LLVM lowers this memset to a libcall, which BPF programs do not support. BPF cannot link against libc and the BPF backend has no `memset` symbol to resolve, which causes compilation to fail. The added `zero_buf()` helper avoids this by writing the byte loop explicitly in a `__noinline` subprogram, so it lands as a single BPF-to-BPF call, and by using `volatile` writes so LLVM's loop-idiom recognition does not re-lower it back into `__builtin_memset`.

### DIFF
--- a/examples/c/bootstrap.bpf.c
+++ b/examples/c/bootstrap.bpf.c
@@ -45,6 +45,9 @@ int handle_exec(struct trace_event_raw_sched_process_exec *ctx)
 	if (!e)
 		return 0;
 
+	/* zero out buffer to avoid leaking prior record bytes */
+	__builtin_memset(e, 0, sizeof(*e));
+
 	/* fill out the sample with data */
 	task = (struct task_struct *)bpf_get_current_task();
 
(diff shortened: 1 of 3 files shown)


anakryiko:
this is no "leak", this data is not privileged and is always readable by user space, there is nothing to fix

